PR #1, opened 2022-07-03
last change round sip

nivBlox:


lior-blox:
@nivBlox Lets add some description on how the sync work for the last change round per signers messages

alonmuroch:
We need to add technical info about what exactly we want to do, just saying we want to save them is not enough.
Also, why is this strictly limited to rc msgs? why not decided as well?

nivBlox:
@alonmuroch 
regarding decided msg's we dont need per signer cause decided holds multi signers.

sips/networking/last_change_round.md
```diff
@@ -0,0 +1,99 @@
...
+Allow nodes to save non committee (validators that are not belong to the node) change round messages in order to serve
+other nodes when requesting last change round messages.
+
+Rational & Design Goals  
+When using subnets instead of topic per validator there is a need for supporting get last change round from any peer in
+the subnet. in order to supports that, each node (peer) needs to handle incoming change round messages and save it per
+signer and only the highest once. the reason for saving messages per signer is to keep the same behavior as it was with
+validator topic.
+
+In order to achieve change round quorum we need 3 different signers. other option is f+1 where we need 2 signers.
```
alonmuroch, line 15: better not set hard coded numbers but rather say 2f+1/ f+1

sips/networking/last_change_round.md
```diff
@@ -0,0 +1,99 @@
...
+*Saving*
+Each operator peer listen for the non committee change round messages within it's subscribed subnets.
+once receive non committee change round message, the following process run -  
+
+- When local not exist. New message saves instantly.
+- New change round message should be with higher or equal height than the local message.
+  If higher, save instantly. if not, check if round is higher. if so, update.
+- need to maintenance the last change round message for each of the signers in the specific validator committee.
+
+```go
```
alonmuroch, line 33: If you need code changes better put it in the spec no? 
SIPs talk more about spec than implementation

sips/networking/last_change_round.md
```diff
@@ -0,0 +1,99 @@
+| Author     | Title                                             | Category | Status |
+|------------|---------------------------------------------------|----------|--------|
+| Niv Muroch | Handle non committee change round for all signers | Core     | Draft  |
```
alonmuroch, line 3: why not add decided msgs as well? isn't it the same thing in terms of spec?

alonmuroch (changes requested):

sips/networking/last_change_round.md
nivBlox: sync change round is not exist in spec. we can exclude this implementation explain from SIP. WDYT?

sips/networking/last_change_round.md
nivBlox: decided msg includes more than 1 signer. so no need to save by signers it is enough as is

alonmuroch:
@nivBlox @lior-blox if we are changing network topology to groups do we need this SIP?

nivBlox:
> 

@alonmuroch @lior-blox 
i think we still need it as we want to make sure getting change round from more than one signer.

alonmuroch:
@nivBlox but this SIP is for non committee validators which we might not have going forward

alonmuroch:
closed, not relevant

PR #2, opened 2022-07-10
Peers Limiting and Balancing

amir-blox:


lior-blox:
> @amir-blox should we start reviewing or it's not final yet?

@alonmuroch it's not final.
maybe we should have a template for how SIP should look like.

alonmuroch:
@amir-blox should we start reviewing ?

amir-blox:
> @amir-blox should we start reviewing ?

@lior-blox @alonmurochwe let's start IMO, we can change/add template afterwards if needed

alonmuroch:
@amir-blox looks good to me, can you:
- make sure we can merge it without conflicts? 
- give it SIP # 3and add it to the all page and core page?
then I'll merge

amir-blox:
Closing this SIP, and adding its content to network spec as it was already implemented.

https://github.com/bloxapp/ssv-spec/blob/network-docs/p2p/SPEC.md#peers-balancing

PR #5, opened 2022-07-26
msg struct and encoding SIP

alonmuroch:
This SIP details new message structs and encoding for QBTF and SSV


lior-blox:
We should include in the SIP the whole ssv message struct, in addition adding encode(ssz) and compression algorithm(snappy/s2)

PR #6, opened 2022-08-01
Compression benchmark comparison

olegshmuelov:


sips/core/msg_struct_encoding.md
```diff
@@ -35,6 +35,52 @@ SSZ is 66% smaller (in the test, 1,640 bytes vs 3,700) vs json, using snappy com
...
+***Benchmark***
+```
+go test -bench=. ./ssz_encoding/qbft -benchmem -benchtime 15s -cpu 2
+
+goos: darwin
+goarch: amd64
+pkg: ssv-experiments/ssz_encoding/qbft
+cpu: Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
+--------------------------------------------------------------------------------------------------------------------------
+NAME-Number of cores                            Iterations           Nanoseconds/op        Number of bytes     Number of alloccs
```
amir-blox, line 65: @olegshmuelov can you also benchmark plain JSON here? as we need to compare to current baseline (JSON)

nivBlox (approved):

sips/core/msg_struct_encoding.md
olegshmuelov: the benchmark comparison for the JSON is in the table

amir-blox: I mean w/o compression (like we have today)

amir-blox (approved):

lior-blox:
@olegshmuelov lets please move the compression and benchmark tests to the bottom

alonmuroch:
@olegshmuelov is this an SIP or? 
If so could you please write it with the same structure?

sips/msg_struct_encoding.md
```diff
@@ -37,6 +37,12 @@ SSZ is 66% smaller (in the test, 1,640 bytes vs 3,700) vs json, using snappy com
 
 SSZ hash tree root enables substituting full data objects with their hash tree root without changing the full message hash tree root which also doesn’t change the BLS signature signing the object. Below we explain how this property is used.
 
+**Compression**  
+We’ve chosen to use [S2](https://github.com/klauspost/compress/tree/master/s2#snappy-compatibility) compression for SSZ encoding, based on the benchmark described below.
```
alonmuroch, line 41: Are we talking about 10% diff between snappy and s2? For interoperability sake isn’t it worth staying with snappy?
How well tested and used s2 is? Who is behind it?

olegshmuelov: I have a dedicated task for this.
Once the spec will be aligned with the new message structs, we can benchmark the real data and have a proper results & decisions.

alonmuroch: I mean the diff between s2 and snappy?

alonmuroch:
@lior-blox closing?

PR #10, opened 2022-09-04
qbft sync triggers

alonmuroch:
SIP covering when a QBFT controller should sync

### DIFF
--- a/sips/qbft_sync.md
+++ b/sips/qbft_sync.md
@@ -37,28 +37,53 @@ func (c *Controller) ProcessMsg(msg *SignedMessage) (bool, []byte, error) {
 	// code after
 }
 
-func (c *Controller) processHigherHeightMsg(msg *SignedMessage) error {
-    added, err := c.HigherReceivedMessages.AddIfDoesntExist(msg)
-    if err != nil {
-        return errors.Wrap(err, "could not add higher height msg")
+func (c *Controller) processFutureMsg(msg *SignedMessage) (*SignedMessage, error) {
+    if c.isDecidedMsg(msg) {
+        return c.uponFutureDecided(msg)
     }
-    if added && c.f1SyncTrigger() {
-        return c.network.SyncHighestDecided(c.Identifier)
+
+    if err := c.verifyAndAddHigherHeightMsg(msg); err != nil {
+        return nil, errors.Wrap(err, "failed adding higher height msg")
     }
+    if c.f1SyncTrigger() {
+        return nil, c.network.SyncHighestDecided(c.Identifier)
+    }
+    return nil, nil
+}
+
+// verifyAndAddHigherHeightMsg verifies msg, cleanup queue and adds the message if unique signer
+func (c *Controller) verifyAndAddHigherHeightMsg(msg *SignedMessage) error {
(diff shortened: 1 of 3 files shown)


sips/qbft_sync.md
```diff
@@ -0,0 +1,106 @@
...
+There are 2 types of syncing: 
+1) Inter-instance syncing - syncing between instances, means syncing a nodes with all decided instances from some past instance height to the highest known decided instance 
+2) Intra-instance syncing - syncing a node, in a specific instance which has yet to decide, with in-instance messages
+
+**Why do we need syncing**  
+Nodes might not receive all messages for various reasons (being offline, disconnected, etc) which can result in a node not receiving up-to-date decided messages (left behind on an old height) and/ or not receiving in-instance messages which might cause the entire committee to go into round changes.  
+Round changes in QBFT are exponential to enable un-synced nodes to ultimately catch up. If a committee gets into high change rounds (timeouts can be hours and even days) it can take a significant amount of time for the committee to "return" back to executing duties.  
+Introducing inter/ intra instance syncing can cut that time significantly and make the committee more lively.
+
+_Example_: only 2 nodes are online starting instance at height H, after some time they are both on round R in which the round timeout is 2^R seconds.
```
amir-blox, line 17: timeout is 3^R

sips/qbft_sync.md
```diff
@@ -0,0 +1,106 @@
...
+Most times when a committee is "stuck" it's because it's waiting for such a quorum. Syncing directly the highest round change will enable the syncing node to quickly form a quorum and/ or speef up F+1 sync.
+
+_**Continuous get highest round change**_  
+Due to the importance of getting the most recent rc message we employ a continuous get highest round change in certain conditions.    
+The above means that a get highest change round call will be triggered every X^(R-4) until the instance decides.
+
+Params:  
+Round Timeout(sec): X^R, R >= 0
+Next get highest round change(sec): X^(R-4) ∀ R>6
+Next get highest round frequency (within each round timeout) = X^R/X^(R-4)  = X^4
```
amir-blox, line 96: does frequency means rate?

sips/qbft_sync.md
```diff
@@ -0,0 +1,106 @@
...
+    })
+}
+```
+
+This sync only fetches the highest round change because for every round > FirstRound the following proposal requires a quorum of round change messages.  
+Most times when a committee is "stuck" it's because it's waiting for such a quorum. Syncing directly the highest round change will enable the syncing node to quickly form a quorum and/ or speef up F+1 sync.
+
+_**Continuous get highest round change**_  
+Due to the importance of getting the most recent rc message we employ a continuous get highest round change in certain conditions.    
+The above means that a get highest change round call will be triggered every X^(R-4) until the instance decides.
```
amir-blox, line 91: `X^(R-4)` interval is producing many messages for round 7 (`3^3`)

What about starting to do this from round 8 (`3^4`) instead of 7? or to increase the exponent `X^(R-3)`?

alonmuroch: Sure I can change to R-3

sips/qbft_sync.md
alonmuroch: it's an example

PR #15, opened 2022-09-12
ecies share encryption

alonmuroch:


alonmuroch:
When switching groups there is nothing that guarantees the actual transfer of slashing data, maybe we should limit switching at max f operators so to ensure you always have operators with the historical data? 
@lior-blox

alonmuroch:
ready to merge

PR #20, opened 2022-11-23
MEV

moshe-blox:
This PR is not final and has two cases where we must narrow it down between different options:
- Under "Node behaviour"
- Under "Blinded block proposals"

A couple more TODO comments raise questions as well.

### DIFF
--- a/sips/mev.md
+++ b/sips/mev.md
@@ -1,0 +1,74 @@
+| Author                    | Title                            | Category | Status               |
+|---------------------------|----------------------------------|----------|----------------------|
+| Moshe Revah (@moshe-blox) | Support externally built blocks | Core     | open-for-discussion  |
+
+## Summary
+Support the production and proposal of blocks built by external entities, allowing operators to access blocks from a marketplace of builders such as [Flashbots](https://boost-relay.flashbots.net/).
+
+## Rational & Design Goals
+Block building is a computationally intensive task, and even more so when blockspace utilization and profit maximization are in mind. Additionally, many of the more profitable transactions, so-called MEV, are private and only accessible to validators through external block builders.
+
+Currently, operators wishing to offer competitive returns must be sophisticated block builders with private transaction flows. This SIP aims to change that by allowing them to register with and propose blocks from external builders who comply to [ethereum/builder-specs](https://github.com/ethereum/builder-specs).
+
+## Specification
+
+### MEV-supporting operators
+
+Operators who wish to support blinded blocks must configure their node to do so, otherwise their node will reject signing those.
+
+Validators who want to propose MEV blocks would have to avoid registering with a combination of MEV and non-MEV operators, as the latter would reject any blinded blocks and result in a missed duty.
+
+> TODO: can't MEV-supporting leaders just fall back to non-blinded blocks if at least one of the other operators is non-MEV supporting? This should allow such combinations.
+
+### Validator registration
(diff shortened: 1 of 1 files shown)


sips/mev.md
```diff
@@ -0,0 +1,111 @@
...
+
+    Unrealistic?
+
+#### Secnario B — Block is published by all operators
+
+1. Leader produces a `BlindedBeaconBlock` from their Beacon node at [/eth/v1/validator/blinded_blocks](https://ethereum.github.io/beacon-APIs/#/Validator/produceBlindedBlock) and broadcasts it together with the endpoint of it's builder as `BuilderBlindedBeaconBlock`.
+
+    ```go
+    type BuilderBlindedBeaconBlock {
+        BuilderEndpoint    string
```
alonmuroch, line 85: Is it the builder or relay endpoint ?

alonmuroch (changes requested):

sips/mev.md
alonmuroch: also seems super easily abused

alonmuroch (changes requested):

sips/mev.md
moshe-blox: > Is it the builder or relay endpoint ?

I call it "builder" instead of "relay" because both relays and individual builders implement the [builder-specs](https://github.com/ethereum/builder-specs), so theoretically you should be able to point your Beacon node to an individual builder (if you have some privileged access to them?) instead of a relay.

> also seems super easily abused

Yeah, operators trust the leader on this one :\

moshe-blox:
> @moshe-blox not sure entirely what is suggested in the SIP. Do we switch to have all blocks be blinded blocks? (MEV and non)? If so how do committee members get the full transaction list post signature (for normal blocks)?

Currently this SIP only adds support for blinded blocks and assumes we still propose non-blinded as usual. Additionally, it suggests to even enforce non-blinded blocks in-protocol for operators who choose to be "non-MEV" as discussed under "Node behaviour -> Hard restriction".

If we choose to support non-blinded blocks, then yeah we'd have to transmit the entire block with the full transactions in SSV, so that all operators can publish it to their Beacon node.

alonmuroch:
@moshe-blox I think we should try to have just 1 flow (blinded blocks) for both cases. 
We can maybe consider to simplify things by making only the leader broadcast the block which will remove the need for the others to recover the contents.
It does open it up to rogue leader attacks

sips/mev.md
```diff
@@ -0,0 +1,111 @@
...
+
+2. Operators sign the `BlindedBeaconBlock`.
+
+3. Operators submit the `SignedBlindedBeaconBlock` to the builder's endpoint at [/eth/v1/builder/blinded_blocks](https://ethereum.github.io/builder-specs/#/Builder/submitBlindedBlock) to reveal the transactions and unblind the block into a `SignedBlindedBlock`.
+
+4. Operators publish the `SignedBeaconBlock` to their Beacon node at [/eth/v1/beacon/blocks](https://ethereum.github.io/beacon-APIs/#/Beacon/publishBlock).
+
+    ##### Issues
+
+    ###### Unknown builder
```
alonmuroch, line 98: This means that blinded blocks are basically only broadcastable by the leader...

alonmuroch (changes requested):

PR #25, opened 2023-05-10
MEV SIP (v2)

moshe-blox:


sips/mev.md
```diff
@@ -0,0 +1,87 @@
...
+
+#### Signing
+
+At the start of every slot, operators select their active validators with `ShouldRegisterValidatorAtSlot`, and for those, produce a [`builder-specs/SignedValidatorRegistration`](https://ethereum.github.io/builder-specs/#model-SignedValidatorRegistration) with their preferred `fee_recipient`.
+
+```go
+ValidatorRegistrationSlotInterval = 16 * SlotsPerEpoch
+
+func ShouldRegisterValidatorAtSlot(index phase0.ValidatorIndex, slot phase0.Slot) bool {
+    return (index + slot) % ValidatorRegistrationSlotInterval == 0
```
alonmuroch, line 41: why is validator index important here?

sips/mev.md
```diff
@@ -0,0 +1,87 @@
...
+>
+> 1. Builder doesn't build a block, and Beacon node falls back to locally-built block from it's execution layer.
+> 2. Builder doesn't reward the validator's `fee_recipient` because it isn't aware of it yet.
+> 3. Builder rewards a potentially different `fee_recipient` from the validator's latest registration (such as a registration prior to onboarding to SSV.)
+
+#### Issue: Gas limits
+
+Unlike standard validator clients, gas limits are not set by validators, but rather by their operators.
+
+This SIP proposes to hardcode the gas limit to 30 million (which is the default in Prysm and Lighthouse), but recommends to keep watching it and modify if necessary.
```
alonmuroch, line 61: set where?

sips/mev.md
```diff
@@ -0,0 +1,87 @@
...
+
+When a validator has a proposal duty, their operators:
+
+1. Produce a `BlindedBeaconBlock` from their Beacon node at [/eth/v1/validator/blinded_blocks](https://ethereum.github.io/beacon-APIs/#/Validator/produceBlindedBlock)
+2. Reach consensus on and sign the round leader's `BlindedBeaconBlock`
+3. Submit the `SignedBlindedBeaconBlock` to their Beacon nodes at [/eth/v1/beacon/blinded_blocks](https://ethereum.github.io/beacon-APIs/#/Beacon/publishBlindedBlock)
+
+Under the hood, each operator's Beacon node would attempt to unblind the block from it's connected builder(s) by revealing the signature to them, which would fail if they haven't built or don't know of this block.
+
+Therefore, for unblinding to succeed, operators' Beacon nodes must share the same builder(s) with the round leader's Beacon node.
```
alonmuroch, line 73: same builder or relay? relay hold the full block no?

sips/mev.md
```diff
@@ -0,0 +1,87 @@
...
+
+Since SSV can't enforce a specific set of builder(s) at the protocol-level, operators would have to reach social consensus on which builder(s) to configure their Beacon nodes with.
```
alonmuroch, line 75: what happens if there is a mix of both ssv operators proposing blinded and non blinded blocks in the same cluster?

alonmuroch (changes requested):

sips/mev.md
moshe-blox: That's just used to reduce the CPU/network load by distributing the signing of `ValidatorRegistration` across many slots (10 epochs in the [most recent commit](https://github.com/bloxapp/SIPs/blob/b01c2885c0fd1a9a17e75656ad5f891892166c7b/sips/mev.md#signing)), so each validator would have a registration duty scheduled at a deterministic slot.

We can use the public key instead, which would allow to register non-active validators to MEV builders (although I'm unsure if it's even possible).

sips/mev.md
moshe-blox: In validator clients, it's configurable with a default of 30 million. In SSV it's currently hardcoded to the same default: https://github.com/bloxapp/ssv/blob/38193908b145dbe29186de34a6c60df1085c92e4/beacon/goclient/proposer.go#L27

moshe-blox: Do you think it should be configurable in SSV? @alonmuroch

sips/mev.md
moshe-blox: Currently nodes with blinded proposals disabled are not rejecting blinded blocks.

sips/mev.md
alonmuroch: That's ethereum dependent so not sure

sips/mev.md
alonmuroch: and viceversa?

lior-blox: I think we discussed it should be strict.
If Im an operator that doesn't want to support MEV, I should not participate in the Consensus if the leader proposed MEV

lior-blox: viceversa should be allowed

moshe-blox: My bad, I thought we dropped this, but the implementation actually rejects QBFT proposals with blinded blocks if you're an operator with blinded disabled. Fix the SIP...

Vice-versa is allowed, yeah.

sips/mev.md
moshe-blox: We can let operators configure this, but if they differ, they will not be able to reconstruct signatures for `ValidatorRegistration` because it contains this value.

Social consensus on gas limit hasn't changed (at least) since The Merge, and if it does we'd want operators to switch at roughly the same time to avoid  differing `ValidatorRegistration`, so maybe waiting for this to change and then releasing a version with a transition scheduled at a specific epoch/slot is safer?

moshe-blox: Updated: https://github.com/bloxapp/SIPs/pull/25/files#diff-ffa7094aebeadab838e68f9af7d1be31bb06e525d5b9436cf8732dff04ab62bbR17-R19

sips/mev.md
moshe-blox: Updated: https://github.com/bloxapp/SIPs/pull/25/files#diff-ffa7094aebeadab838e68f9af7d1be31bb06e525d5b9436cf8732dff04ab62bbR17-R19

lior-blox (approved):

sips/mev.md
```diff
@@ -0,0 +1,91 @@
...
+
+Currently, operators wishing to offer competitive returns must be sophisticated block builders with private transaction flows. This SIP aims to change that by allowing them to register with and propose blocks from external builders who comply to [ethereum/builder-specs](https://github.com/ethereum/builder-specs).
+
+## Specification
+
+### Configuration
+
+Operators who wish to propose blinded blocks must configure their node to do so, otherwise it would only propose standard blocks and reject QBFT proposals containing blinded blocks.
+
+Therefore, validators should solely select operators with matching configuration, otherwise consensus might be unreachable.
```
GalRogozinski, line 21: what is the motivation for disabling support?
Why not just always have it on?

If the beacon node will want to create a non-blinded block then it is fine. Else it is also fine.

Btw, the used relay is a configuration of the beacon node only?

sips/mev.md
GalRogozinski: @lior-blox 
Let say in a 4 node cluster 2 operators don't want to support MEV?
So no block will ever be produced?

Also why would some operators oppose to use MEV? I can only think of idealogical reasons or deep mistrust of relays. In this case, I think it is more of the validator's choice and not the operator.

sips/mev.md
```diff
@@ -0,0 +1,91 @@
...
+ValidatorRegistrationSlotInterval = 10 * SlotsPerEpoch
+
+func ShouldRegisterValidatorAtSlot(index phase0.ValidatorIndex, slot phase0.Slot) bool {
+    return (index + slot) % ValidatorRegistrationSlotInterval == 0
+}
+```
+
+#### Publishing
+
+Every slot, operators should publish the most recent successfully signed registration for `1/SlotsPerEpoch` of their validators, so that all validators are registered at least once per epoch.
```
GalRogozinski, line 49: So let say that the most recent signed registration is at slot X, then operators will publish the same registration at X+1, X+2 and so on?

Is it neccessary?

moshe-blox: Regardless of when was the last signed registration, you submit for each validator once an epoch (the next portion of 1/32 at every slot). This SIP suggests to split the submissions per-slot, instead of a big one at the start/end of the epoch. Do you think I should specify exactly how to split your validators per-slot in this SIP? (for example with `(validatorIndex + slot) % 32 == 0`)

Why to re-submit the same signature?
1. A new relay may have been added and hasn't seen your registration yet
2. An existing relay forgot your registration for whatever reason
3. The Beacon node **might** be programmed to use the fee recipient from your registration when falling back to build local blinded blocks, but it forgets your registration after restarts

Prysm & Lighthouse seem to only sign once (and never again until preferences changed or node restarts), and publish once at the start of every epoch.

Ideally, we'd schedule signatures around `ValidatorAddedEvent` and `FeeRecipientAddressUpdated` to avoid re-signing when nothing has changed, or first signing too late (after a block proposal). However, we don't think it's too bad, and would like to keep that for a future PR.

sips/mev.md
moshe-blox: Beacon nodes would always return a blinded block if you asked for it. If it can't source it externally, it would build it locally.

I see where the confusion is, committed this to clarify:
> Operators who wish to propose blinded blocks must configure their **SSV** node to do so

sips/mev.md
lior-blox: @GalRogozinski 
In the scenario, only proposals that the leader is non mev will be successful.
It could be various reasons like ideology or legal etc.

The operators will use metadata to let the validators If they MEV support and which relays

sips/mev.md
```diff
@@ -0,0 +1,91 @@
...
+
+Since SSV can't enforce a specific set of builder(s) at the protocol-level, operators would have to reach social consensus on which builder(s) to configure their Beacon nodes with.
+
+#### Locally-built blinded blocks
+
+Sometimes, a Beacon node falls back to a locally-built `BlindedBeaconBlock` from it's execution layer, either because it's configured to do so under certain conditions (profitability threshold; chain is unstable) or because the builder(s) are unavailable.
+
+A locally-built blinded block can only be unblinded by the same Beacon node which built it, which unfortunately means that only the round leader would successfully submit the block, thereby reducing the total network outreach of the proposal down to a single Beacon node.
+
+We should consider pushing operators to configure their node for less harsh fallback conditions, so that more of their blocks are externally-built and can be successfully submitted by other operators' Beacon nodes.
```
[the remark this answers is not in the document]
moshe-blox: Yeah, from the paragraph above this:
> either because it's configured to do so under certain conditions (profitability threshold; chain is unstable) or because the builder(s) are unavailable.

sips/mev.md
moshe-blox: Rewritten this:

> Every epoch, operators should publish the most recent successfully signed registration for their entire validator set.
>
> This SIP suggests to spread registrations between slots. For example, at every slot, operators may publish for a fraction of their validators (corresponding to `1/SlotsPerEpoch`), ensuring a more evenly distributed load on the Beacon node and builder(s).

sips/mev.md
```diff
@@ -0,0 +1,93 @@
...
+
+Builders require validators to publish a [`builder-specs/SignedValidatorRegistration`](https://ethereum.github.io/builder-specs/#model-SignedValidatorRegistration) to set their `fee_recipient` and `gas_limit` preferences.
+
+When building a block for a validator, builders refer only to the registration with the highest timestamp.
+
+In the wild, some Ethereum validator clients currently publish this message every epoch.
+
+#### Fee recipients
+
+Validators may set their preferred `fee_receipient` address by calling `setFeeRecipientAddress` in the `SSVNetwork` contract. Validators may repeat this call as their preference changes over time.
```
nkryuchkov, line 33:
```suggestion
Validators may set their preferred `fee_recipient` address by calling `setFeeRecipientAddress` in the `SSVNetwork` contract. Validators may repeat this call as their preference changes over time.
```

sips/mev.md
```diff
@@ -0,0 +1,93 @@
...
+func ShouldRegisterValidatorAtSlot(index phase0.ValidatorIndex, slot phase0.Slot) bool {
+    return (index + slot) % ValidatorRegistrationSlotInterval == 0
+}
+```
+
+#### Publishing
+
+Every epoch, operators should publish the most recent successfully signed registration for their entire validator set.
+
+This SIP suggests to spread registrations between slots. For example, at every slot, operators may publish only for a fraction of their validators (corresponding to `1/SlotsPerEpoch`), ensuring a more evenly distributed load on the Beacon node and builder(s).
```
nkryuchkov, line 51: Should we add a note about batching (sending 1/32 validators each slot might hit rate limits) and about sending also cached registrations every epoch?

moshe-blox: However the implementation decides to "publish the most recent successfully signed registration" is their call, the SIP doesn't mind.

Whether it's cache/database and whether to publish every X slots or once an epoch are implementation details, especially because figuring out the ideal rate at which to "spam" your mev-boost requires experimentation and tuning in a real-world environment.

moshe-blox: This is why the SIP only "suggests" and then provides an "example", to inspire the implementation to consider and experiment with these ideas.

alonmuroch (changes requested):

sips/mev.md
```diff
@@ -0,0 +1,103 @@
...
+```solidity
+function setFeeRecipientAddress(address feeRecipientAddress) external;
+
+event FeeRecipientAddressUpdated(address indexed owner, address recipientAddress);
+```
+
+Validators may repeat this call as their preference changes over time.
+
+Operators should incorporate the `fee_recipient` address from the most recent event emitted by a specific `owner` when constructing `ValidatorRegistration` messages for validators associated with that `owner`.
+
```
GalRogozinski, line 44: This is also how the general registration process is initiated by the user (validator)?
If so, let's make it clear?
If not? then how it is started?

GalRogozinski (changes requested):
There are a few holes in order for the reader to get a good grasp of how this thing should work on SSV.

I would also like to have explanations on what validations we are doing (or not doing) to message registration and why it is fine. This will force us to think about any security issues.

GalRogozinski (changes requested):
There are a few holes in order for the reader to get a good grasp of how this thing should work on SSV.

I would also like to have explanations on what validations we are doing (or not doing) to message registration and why it is fine. This will force us to think about any security issues.

sips/mev.md
```diff
@@ -0,0 +1,93 @@
...
+Block building is a computationally intensive task, and even more so when blockspace utilization and profit maximization are in mind. Additionally, many of the more profitable transactions, so-called MEV, are private and only accessible to validators through external block builders.
+
+Currently, operators wishing to offer competitive returns must be sophisticated block builders with private transaction flows. This SIP aims to change that by allowing them to register with and propose blocks from external builders who comply to [ethereum/builder-specs](https://github.com/ethereum/builder-specs).
+
+## Specification
+
+### Configuration
+
+Operators who wish to propose blinded blocks must configure their SSV node to do so, otherwise it would only propose standard blocks and reject QBFT proposals containing blinded blocks.
```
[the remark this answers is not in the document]
moshe-blox: Yeah, originally I wanted blinded/non-blinded to be a property operators set in the contract with a cooldown period to make them commit to their claim.

So for example if an operator switches from non-blinded to blinded in the contract, but also switches it in his node before the cooldown period is over, his blocks would be rejected by other operators.

Eventually, we didn't go this way, so now rejecting blocks sounds less appealing.

Still, I can think of a few reasons for rejections:
- Regulatory compliance: an operator who wishes to be OPAC-compliant, may want to avoid signing blinded blocks because he can't see whether they contain non-compliant transactions.
- Relay/operator theft: imagine a scenario where many relays/operators suddenly stop paying the fee & MEV rewards to validators. Validators start panicking and migrating to non-blinded operators, and because we verify the fee recipient in non-blinded blocks, they should be safe. However, if non-blinded operators don't reject blinded blocks, there is nothing to prevent 'non-blinded' operators from spontaneously proposing a single blinded block that steals fees, every once in a while.

sips/mev.md
GalRogozinski: This discussion can continue here:
https://github.com/bloxapp/SIPs/pull/25#discussion_r1196242791

sips/mev.md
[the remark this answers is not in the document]
moshe-blox: @GalRogozinski Cooldown period is a commitment between operators to validators. When validators choose a non-blinded operator, they can rest assured that it would not suddenly change it's mind and start proposing blinded blocks. They would know about this change a week (or however long the cooldown period is) in advance.

sips/mev.md
```diff
@@ -0,0 +1,103 @@
...
+event FeeRecipientAddressUpdated(address indexed owner, address recipientAddress);
+```
+
+Validators may repeat this call as their preference changes over time.
+
+Operators should incorporate the `fee_recipient` address from the most recent event emitted by a specific `owner` when constructing `ValidatorRegistration` messages for validators associated with that `owner`.
+
+#### Signing
+
+At the start of every slot, operators select their active validators with `ShouldRegisterValidatorAtSlot`, and for those, produce a [`builder-specs/SignedValidatorRegistration`](https://ethereum.github.io/builder-specs/#model-SignedValidatorRegistration) with their preferred `fee_recipient`.
```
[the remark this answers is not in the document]
moshe-blox: committed, thanks 🙏

sips/mev.md
```diff
@@ -0,0 +1,93 @@
...
+
+### Configuration
+
+Operators who wish to propose blinded blocks must configure their SSV node to do so, otherwise it would only propose standard blocks and reject QBFT proposals containing blinded blocks.
+
+Therefore, validators should solely select operators with matching configuration, otherwise consensus might be unreachable.
+
+### Validator registration
+
+Builders require validators to publish a [`builder-specs/SignedValidatorRegistration`](https://ethereum.github.io/builder-specs/#model-SignedValidatorRegistration) to set their `fee_recipient` and `gas_limit` preferences.
```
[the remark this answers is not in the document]
moshe-blox: [This section specifies when to publish](https://github.com/bloxapp/SIPs/pull/25/files#diff-ffa7094aebeadab838e68f9af7d1be31bb06e525d5b9436cf8732dff04ab62bbR79-R83).

There is no rule in the Ethereum builder-specs regarding when/how often this should be publish, but @lior-blox suggested that we should specify it in the spec so that future SSV implementations would be aligned on it.

sips/mev.md
GalRogozinski: I meant that maybe you add an explanation like you did here https://github.com/bloxapp/SIPs/pull/25#discussion_r1196242791.

If not in the SIP I think it should be written somewhere accessible.

alonmuroch:
@moshe-blox @lior-blox update?

GalRogozinski (approved):

sips/mev.md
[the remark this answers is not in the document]
y0sher: Chiming in here to mention that beacon API specs has changed since this SIP, the separate API calls to the beacon node about getting a full or blinded block have been deprecated, the new V3 api call will get you a blinded block if your beacon node is connected to MEV relays and possible and the blinded block is more profitable than a local one, otherwise it will get you a local full block.
this means its not straight forward anymore to config the SSV operator node to support or not support MEV, this responsibility is now on the beacon node.

PR #26, opened 2023-05-14
updated fork support

alonmuroch:
- updated fork support to ssv-spec main


GalRogozinski (commented):
You wrote "Consider weak subjectivity".
This means you want to add checkpoints?

GalRogozinski (commented):
What bothers me is that unlike ethereum, the consensus participating nodes of an instance is very small.

Meaning that one unupgraded node can have dire consequences.

Maybe this is outside the scope of the SIP, but I am assuming that a good time before the trigger there will be a way for validators to see who upgraded?

GalRogozinski:
@moshe-blox - we should opt for soft forks when possible to reduce code complexity.
@MatheusFranco99 - we should write down a clear procedure for activation epoch. We should also write down when we stop supporting old forks.

PR #28, opened 2023-06-15
Remove QBFT decided instance requirement

alonmuroch:
Removes the requirement for an instance to decided to start the next one

### DIFF
--- a/sips/instance_decided_enforcement.md
+++ b/sips/instance_decided_enforcement.md
@@ -11,7 +11,7 @@ For a new instance to start, the QBFT controller checks that the previous one ha
 Originally this was made so instance heights are predictable, synchronized and easy to sync by every node in the network.
 
 The main reasons to drop these requirements are:
-1) Instance Height can be the duty's epoch number as every duty, at most, there is 1 duty per type of duty per validator.
+1) Instance Height can be the duty's slot number as every duty, at most, there is 1 duty per type of duty per validator.
 2) Syncing historical decided instances can be made easy with the use of epochs, e.g. fetch decided instances for epochs 100-150.
 3) Stalled/ non decided duties have a life expectancy of, at most, 2 epochs, since past that even if executed will not gain rewards for validator (per beacon roles)
 4) There is a difficult balance between efficient constant time timeouts and liveness. There are some complex solutions but since duties have an effective expiration time and are "non-dependent" we can terminate at a certain round.
@@ -22,6 +22,7 @@ The main reasons to drop these requirements are:
 - Remove function CanStartInstance
 - Remove function canStartInstanceForValue
 - Remove function bumpHeight
+- Remove future message processing - no need as each instance is independent. If a future decided message is received it will trigger decided sync
 ```go
 // StartNewInstance will start a new QBFT instance, if can't will return error
 func (c *Controller) StartNewInstance(height Height, value []byte) error {
@@ -39,16 +40,44 @@ func (c *Controller) StartNewInstance(height Height, value []byte) error {
 
 	c.Height = height
 
+    c.forceStopAllInstanceExceptCurrent()
+
 	return nil
 }
+
+func (c *Controller) forceStopAllInstanceExceptCurrent() {
+    for _, i := range c.StoredInstances {
+        if i.State.Height != c.Height {
+            i.ForceStop()
+        }
+    }
+}
+
 ```
 
 **QBFT Instance**
 ```go
+type Instance struct {
+	// ...
+    // forceStop will force stop the instance if set to true
+    forceStop  bool
+}
+
+func (i *Instance) ForceStop() {
+    i.forceStop = true
+}
+
+
+// CanProcessMessages will return true if instance can process messages
+func (i *Instance) CanProcessMessages() bool {
+	return i.State.Round < CutoffRound && !i.forceStop
+}
+```
+```go
 // ProcessMsg processes a new QBFT msg, returns non nil error on msg processing error
 func (i *Instance) ProcessMsg(msg *SignedMessage) (decided bool, decidedValue []byte, aggregatedCommit *SignedMessage, err error) {
-    if i.State.Round == CutoffRound {
-        return false, nil, nil, errors.New("round > cutoff round")
+    if i.CanProcessMessages() {
+        return false, nil, nil, errors.New("instance stopped processing messages")
     }
 	
 	// ...
@@ -59,9 +88,9 @@ func (i *Instance) ProcessMsg(msg *SignedMessage) (decided bool, decidedValue []
 const CutoffRound = 20
 
 func (i *Instance) UponRoundTimeout() error {
-	if i.State.Round == CutoffRound {
-		return errors.New("round > cutoff round")
-	}
+    if i.CanProcessMessages() {
+        return errors.New("instance stopped processing timeouts")
+    }
 
 	newRound := i.State.Round + 1
 	defer func() {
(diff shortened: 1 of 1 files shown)


sips/instance_decided_enforcement.md
```diff
@@ -0,0 +1,97 @@
...
+	// ...
+}
+```
+```go
+// CutoffRound which round the instance should stop its timer and progress no further
+const CutoffRound = 20
+
+func (i *Instance) UponRoundTimeout() error {
+	if i.State.Round == CutoffRound {
+		return errors.New("round > cutoff round")
```
GalRogozinski, line 63: Even though `==` is enough I think `>=` is safer.

```suggestion
	if i.State.Round >= CutoffRound {
		return errors.Errorf("round %d >= cutoff round %d", i.State.Round, CutoffRound)
```

sips/instance_decided_enforcement.md
```diff
@@ -0,0 +1,97 @@
...
+		return errors.Wrap(err, "failed to broadcast round change message")
+	}
+
+	return nil
+}
+```
+
+**Runner**
+- Remove function canStartNewDuty
+- Start new instance on decide(runner Runner, input *types.ConsensusData) error will start the instance with the epoch calculated from the duty slot
```
GalRogozinski, line 88: didn't understand this line... no code for it?

GalRogozinski (changes requested):
Should we also finish the runner? As long as it is running and not decided attackers can send messages and drain resources

sips/instance_decided_enforcement.md
```diff
@@ -0,0 +1,97 @@
...
+**Summary** 
+Describes the remove of requiring a QBFT instance to decided to start the next one. 
+
+**Rational**  
+Currently QBFT instances have an incremental index for their height, starting with 0.
+For a new instance to start, the QBFT controller checks that the previous one has decided.
+Originally this was made so instance heights are predictable, synchronized and easy to sync by every node in the network.
+
+The main reasons to drop these requirements are:
+1) Instance Height can be the duty's epoch number as every duty, at most, there is 1 duty per type of duty per validator.
```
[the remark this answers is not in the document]
alonmuroch: You are right, due to sync committee we need to use slots not epoch

sips/instance_decided_enforcement.md
alonmuroch: ok, please change

sips/instance_decided_enforcement.md
alonmuroch: Start new instance on decide, will start the instance with the duty.slot

alonmuroch:
@GalRogozinski when new instance starts, force start exiting instance

alonmuroch:
@GalRogozinski when new instance starts, force start exiting instance

alonmuroch:
@GalRogozinski @lior-blox sync needs to be changed to epochs/ slots and not by decided range

alonmuroch:
@GalRogozinski @lior-blox we can drop non committee processing + sync highest decided since we can just sync by epoch/ slot

sips/instance_decided_enforcement.md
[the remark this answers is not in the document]
GalRogozinski: just sync committees?
I think also a validator can propose blocks in multiple slots in a single epoch if I am not mistaken?

GalRogozinski:
Maybe for sync committee duties we can terminate faster? Since if they miss their slot then no reward?

sips/instance_decided_enforcement.md
```diff
@@ -0,0 +1,134 @@
...
+// CanProcessMessages will return true if instance can process messages
+func (i *Instance) CanProcessMessages() bool {
+	return i.State.Round < CutoffRound && !i.forceStop
+}
+```
+```go
+// ProcessMsg processes a new QBFT msg, returns non nil error on msg processing error
+func (i *Instance) ProcessMsg(msg *SignedMessage) (decided bool, decidedValue []byte, aggregatedCommit *SignedMessage, err error) {
+    if i.CanProcessMessages() {
+        return false, nil, nil, errors.New("instance stopped processing messages")
```
GalRogozinski, line 80: This error is received by `BaseRunner.baseConsensusMsgProcessing`.
The runner should set its state to `finished` if this happens.

sips/instance_decided_enforcement.md
```diff
@@ -0,0 +1,134 @@
...
+
+	newInstance := c.addAndStoreNewInstance()
+	newInstance.Start(value, height)
+
+	c.Height = height
+
+    c.forceStopAllInstanceExceptCurrent()
+
+	return nil
+}
```
GalRogozinski, line 46: Maybe we should have a check that the new instance's height is larger than any stored instance.
```
if height <= c.Height {
     return errors.New("past height")
```

Imagine a bug in a syncing beacon node that for some weird reason gives the nodes duties out of order.

sips/instance_decided_enforcement.md
```diff
@@ -0,0 +1,134 @@
...
+    if i.CanProcessMessages() {
+        return false, nil, nil, errors.New("instance stopped processing messages")
+    }
+	
+	// ...
+}
+```
+```go
+// CutoffRound which round the instance should stop its timer and progress no further
+const CutoffRound = 20
```
GalRogozinski, line 88: Currently we start slow timeout (2 minutes) at round 9.
Meaning we wait more than 22 minutes in order to stop.

2 epochs in Ethereum are approximately 13 minutes.
I understand we need some safety time maybe... but this looks like a lot.

How was this number chosen?

sips/instance_decided_enforcement.md
alonmuroch: good catch, fixed

sips/instance_decided_enforcement.md
alonmuroch: correct, fixed

sips/instance_decided_enforcement.md
```diff
@@ -0,0 +1,150 @@
...
+~~~
+
+**QBFT Controller**
+- Remove function CanStartInstance
+- Remove function canStartInstanceForValue
+- Remove function bumpHeight
+- Remove future message processing - no need as each instance is independent. If a future decided message is received it will trigger decided sync
+```go
+// StartNewInstance will start a new QBFT instance, if can't will return error
+func (c *Controller) StartNewInstance(height Height, value []byte) error {
```
y0sher, line 43: If height is just slot now, maybe we should just call it slot?

GalRogozinski: this will be a huge change for the code and internal states 😅 
also @alonmuroch wanted to keep it as "height"... maybe to keep our concepts seperate from eth

you do have a point though

sips/instance_decided_enforcement.md
```diff
@@ -0,0 +1,150 @@
...
+	if err := c.GetConfig().GetValueCheckF()(value); err != nil {
+		return errors.Wrap(err, "value invalid")
+	}
+	
+	if height <= c.Height {
+		return errors.New("invalid instance height")
+	}
```
GalRogozinski, line 50: I am just wondering now if this should be on the instance level or on the duty level...
Technically it should be a duty, however we currently have "validator registration" implemented as a duty which may be an issue...

sips/instance_decided_enforcement.md
```diff
@@ -0,0 +1,150 @@
+| Author      | Title                        | Category | Status |
+|-------------|------------------------------|----------|--------|
+| Alon Muroch | Instance Decided Enforcement | Core     | open-for-discussion  |
+
+**Summary**
+
+No longer require a QBFT instance to decide in order to start the next one.
+If an instance is stalled, it will be terminated after a certain round or when a new instance starts.
+The instance's height will be the duty's slot number. This will allow for easier syncing.
```
andrew-blox, line 9: I see some confusion here.  In our logs we will have logs concerning height and some concerning slot.  I feel users will ask what is the difference here as it was my first question.  I understand we want to use "height" instead of slot to stay with proper QBFT terminology but maybe its worth removing "slot" terminology as well.

andrew-blox:
@y0sher Do we save somewhere highest slot ever received?  For ex. i have height of 100 now, 1 sec later got a slot now of 99 will this instance be rejected?

sips/instance_decided_enforcement.md
```diff
@@ -0,0 +1,152 @@
...
+    qbft.Height(input.Duty.Slot),
+    byts,
+); err != nil {
+    return errors.Wrap(err, "could not start new QBFT instance")
+}
+```
+
+**Syncing**  
+- Light nodes do not need to sync highest decided at all
+- Syncing for decided messages is done by epoch range
```
y0sher, line 149: So light nodes still need to sync decided messages?

y0sher (approved):

sips/instance_decided_enforcement.md
GalRogozinski: I think @alonmuroch meant for full nodes?

Let's move discussion about full nodes syncing to https://github.com/bloxapp/ssv-spec/issues/280

GalRogozinski (approved):

lior-blox (approved):

PR #34, opened 2023-08-24
Updated fork support SIP

MatheusFranco99:
Updating fork support SIP due to comments on:
- https://github.com/bloxapp/SIPs/pull/26
- https://github.com/bloxapp/ssv-spec/pull/259


sips/fork_support.md
```diff
@@ -20,56 +20,259 @@ Client implementation should calculate when a new duty starts to decide if it st
...
-    PrimusTestnet  = DomainType{0x0, 0x0, PrimusChain, 0x0}
-    ShifuTestnet   = DomainType{0x0, 0x0, ShifuChain, 0x0}
-    ShifuV2Testnet = DomainType{0x0, 0x0, ShifuChain, 0x1}
-    JatoTestnet    = DomainType{0x0, 0x0, JatoChain, 0x1}
+	GenesisMainnet = DomainType{0x0, 0x0, byte(MainnetNetworkID), 0x0}
+	PrimusTestnet  = DomainType{0x0, 0x0, byte(PrimusNetworkID), 0x0}
+	ShifuTestnet   = DomainType{0x0, 0x0, byte(ShifuNetworkID), 0x0}
+	ShifuV2Testnet = DomainType{0x0, 0x0, byte(ShifuNetworkID), 0x1}
+	JatoTestnet    = DomainType{0x0, 0x0, byte(JatoNetworkID), 0x0}
+	JatoV2Testnet  = DomainType{0x0, 0x0, byte(JatoNetworkID), 0x1}
```
olegshmuelov, line 113: Just for reference
The current(ongoing) [JatoV2 network](https://github.com/bloxapp/ssv/blob/beadd23010b4c8f8bc25cb07d00bd534a70403ef/networkconfig/jato-v2.go#L14) behave as different networkID `0x4`
 ```go
 spectypes.DomainType{0x0, 0x0, 0x4, 0x1}
 ```

sips/fork_support.md
```diff
@@ -20,56 +20,259 @@ Client implementation should calculate when a new duty starts to decide if it st
...
-            Epoch:  0,
-            Domain: GenesisMainnet,
-        },
-    }
+func (networkID NetworkID) DefaultFork() *ForkData {
+	return networkID.GetForksData()[0]
 }
 
+// GetCurrentFork returns the ForkData with highest Epoch smaller or equal to "epoch"
+func (networkID NetworkID) GetCurrentFork(epoch phase0.Epoch) (*ForkData, error) {
```
olegshmuelov, line 157: If we want to use GetCurrentFork name for this method, we should not pass epoch but get current epoch from beacon network
I would suggest to rename it to smtg like `ForkAtEpoch` or `ActiveForkAtEpoch`

sips/fork_support.md
```diff
@@ -20,56 +20,259 @@ Client implementation should calculate when a new duty starts to decide if it st
...
+	if len(forks) == 0 {
+		return nil, errors.New("GetCurrentFork: fork list by GetForksData is empty.")
+	}
+
+	current_fork := forks[0]
+	for _, fork := range forks {
+		if fork.Epoch <= epoch && fork.Epoch > current_fork.Epoch {
+			current_fork = fork
+		}
+	}
```
olegshmuelov, line 171:
```suggestion
	var current_fork *ForkData
	for _, fork := range forks {
		if fork.Epoch <= epoch {
			current_fork = fork
		}
	}
```

olegshmuelov (commented):
have few comments, but in general LGTM

additional concern is to add the beaconNetwork to config that can be accessible from the instance to get the current epoch to build the msgIDs with the right domain type

sips/fork_support.md
MatheusFranco99: Oh okay, thanks!

sips/fork_support.md
MatheusFranco99: Sure, well observed. Thanks!

sips/fork_support.md
MatheusFranco99: So the list of forks is guaranteed to ordered, right?

olegshmuelov: > So the list of forks is guaranteed to ordered, right?

right, we can do the same as beacon API `/eth/v1/config/fork_schedule`
```
{
    "data": [
        {
            "previous_version": "0x00001020",
            "current_version": "0x00001020",
            "epoch": "0"
        },
        {
            "previous_version": "0x00001020",
            "current_version": "0x01001020",
            "epoch": "36660"
        },
        {
            "previous_version": "0x01001020",
            "current_version": "0x02001020",
            "epoch": "112260"
        },
        {
            "previous_version": "0x02001020",
            "current_version": "0x03001020",
            "epoch": "162304"
        }
    ]
}
```

but not sure why we need the previous version

MatheusFranco99: Indeed, I don't see why we should use either. In case we need the previous fork of a certain epoch, we can add a function later to it. What do you think?

olegshmuelov: Sure

sips/fork_support.md
```diff
@@ -20,56 +20,260 @@ Client implementation should calculate when a new duty starts to decide if it st
...
+	ret := sha256.Sum256(byts)
+	return ret[:], nil
+}
+```
+
+### Structures access to types
+
+- `Share`: It's more correct for the structure to hold the information on the network that it's in, instead of the network and fork number.
+    Thus, _Share_ will contain a `NetworkID` field instead of a `DomainType` field.
```
alonmuroch, line 189: Is that so important? domain type contains network

sips/fork_support.md
```diff
@@ -20,56 +20,260 @@ Client implementation should calculate when a new duty starts to decide if it st
...
+	}
+
+	// covers height == 0 case
+	if c.StoredInstances.FindInstance(height) != nil {
+		return errors.New("instance already running")
+	}
+
+    c.config.SetSignatureDomainType(domainType) // <-- updates its config's DomainType
+    c.Identifier = identifier // <-- updates its identifier
+	c.Height = height
```
alonmuroch, line 215: align correctly

sips/fork_support.md
```diff
@@ -20,56 +20,260 @@ Client implementation should calculate when a new duty starts to decide if it st
...
+	if height < c.Height {
+		return errors.New("attempting to start an instance with a past height")
+	}
+
+	// covers height == 0 case
+	if c.StoredInstances.FindInstance(height) != nil {
+		return errors.New("instance already running")
+	}
+
+    c.config.SetSignatureDomainType(domainType) // <-- updates its config's DomainType
```
alonmuroch, line 213: is that a must? config is an interface provided on NewController, can't we just change it on implementation and it will propagate to controller?

sips/fork_support.md
```diff
@@ -20,56 +20,260 @@ Client implementation should calculate when a new duty starts to decide if it st
...
+type BaseRunner struct {
+	State          *State
+	Share          *types.Share
+	QBFTController *qbft.Controller
+	BeaconNetwork  types.BeaconNetwork
+	BeaconRoleType types.BeaconRole
+	highestDecidedSlot spec.Slot
+
+	// Current fork associated to Share.DomainType
+	CurrentFork *types.ForkData  // <-- new field
```
alonmuroch, line 237: We can get form from domain right? why add it?

alonmuroch (changes requested):

sips/fork_support.md
MatheusFranco99: It's not that important.
It's a two-way relation. We can derive NetworkID from DomainType and vice-versa, so in practice Share can have either one.
I thought it would be more correct if the Share was to contain the NetworkID instead of DomainType. But I can roll back this change. What do you think?

sips/fork_support.md
MatheusFranco99: Yes, we can. Updating :)

sips/fork_support.md
MatheusFranco99: The controller is the first structure in the hierarchy to contain the config. Though config is passed as an argument, the runner doesn't have access to it and thus the Controller needs to be the one to update it. What do you think?

sips/fork_support.md
MatheusFranco99: Fixed

MatheusFranco99:
Regarding the JatoDomain having an initial fork byte as 0x1 (similarly for JatoV2). I was talking with @olegshmuelov, and we thought that replacing it with 0 could cause some problems. Though it's inconsistent to start with 0x1, the change:
- turns many spec tests invalid (and, likely, the same will occur in SSV tests)
- could cause signature errors in real nodes
- to replace it now could cause even more confusion regarding this byte
So we thought it would be better to keep it that way.

sips/fork_support.md
GalRogozinski: I would roll back. This is because:

1. I am not sure it is a two way relation. Note that DomainType has 2 TBD bytes. Are you sure they can always be derived.
2. If it can't be derived, if you look at the diagram you made, for "duties" that don't have a qbft Controller we will have no way to derive domain types.

sips/fork_support.md
```diff
@@ -20,56 +20,247 @@ Client implementation should calculate when a new duty starts to decide if it st
...
-        return genesisForks()
-    default:
-        return []*ForkData{}
-    }
+func (networkID NetworkID) GetForksData() []*ForkData {
+	switch networkID {
+	case MainnetNetworkID:
+		return mainnetForks()
+	default:
+		return []*ForkData{
```
GalRogozinski, line 134: This should return a 0 length array so `ForkAtEpoch` would err

sips/fork_support.md
```diff
@@ -20,56 +20,247 @@ Client implementation should calculate when a new duty starts to decide if it st
...
+}
+
+// GetCurrentFork returns the ForkData with highest Epoch smaller or equal to "epoch"
+func (networkID NetworkID) ForkAtEpoch(epoch phase0.Epoch) (*ForkData, error) {
+	// Get list of forks
+	forks := networkID.GetForksData()
+
+	// If empty, raise error
+	if len(forks) == 0 {
+		return nil, errors.New("GetCurrentFork: fork list by GetForksData is empty.")
```
GalRogozinski, line 164:
```suggestion
		return nil, errors.New("GetCurrentFork: fork list by GetForksData is empty. Unknown Network")
```

sips/fork_support.md
MatheusFranco99: I thought that it was a two-way relation because with the NetworkID we can get the forks (GetForksData()) and then get the domain of the current fork.

sips/fork_support.md
```diff
@@ -20,56 +20,247 @@ Client implementation should calculate when a new duty starts to decide if it st
...
+		domainType := b.Share.NetworkID.ForkAtEpoch(currentEpoch)
+		identifier := types.NewMsgID(domainType,b.Share.ValidatorPubKey[:],b.BeaconRoleType)
+		return identifier[:]
+	}
+}
+
+func (b *BaseRunner) GetDomainTypeF() func() types.DomainType {
+	return func() types.DomainType {
+		currentEpoch := b.BeaconNetwork.EstimatedCurrentEpoch()
+		return b.Share.NetworkID.ForkAtEpoch(currentEpoch)
```
GalRogozinski, line 263: This returns DomainType?

GalRogozinski (changes requested):

sips/fork_support.md
GalRogozinski: What about the TBD bytes?

MatheusFranco99: Sorry, I mean that the Fork has the whole Domain data associated with an epoch, not just the fork number. So we get the whole Domain (all 4 bytes) through the GetForksData

sips/fork_support.md
MatheusFranco99: This returns a function that returns the current DomainType

sips/fork_support.md
GalRogozinski: LGTM

It should be updated everytime a new instance starts because a fork may get activated.

sips/fork_support.md
GalRogozinski: What I meant is that 		
```
return b.Share.NetworkID.ForkAtEpoch(currentEpoch)
```

will return forkdata and an error. It won't return the domain type.

MatheusFranco99: Oh sorry, yeah, updating it now

sips/fork_support.md
```diff
@@ -20,56 +20,250 @@ Client implementation should calculate when a new duty starts to decide if it st
...
+	FutureMsgsContainer map[types.OperatorID]Height
+	DomainTypeF         DomainTypeF // <-- replace Domain types.DomainType to DomainType getter function
+	Share               *types.Share
+	config              IConfig
+}
+
+func (c *Controller) StartNewInstance(height Height, value []byte) error {
+	// ...
+
+	c.config.SetSignatureDomainType(c.DomainTypeF()) // <-- updates its config using its new DomainF
```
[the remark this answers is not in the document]
MatheusFranco99: Sure, sorry. Updating it now!
Indeed 😅

GalRogozinski (approved):
LGTM :-)

olegshmuelov (approved):

sips/fork_support.md
```diff
@@ -20,56 +20,254 @@ Client implementation should calculate when a new duty starts to decide if it st
...
+- `Instance`:
+  - Must keep its identifier to send properly formed messages.
+- `Controller`:
+  - Should replace its _Identifier_ with an _IdentifierF_ getter function, passed by the Runner. With the function instead of the fixed value, it can compute the up-to-date identifier when needed.
+  - Since the controller's `config` must also be updated, the controller should also contain a _DomainTypeF_ function, instead of its fixed _DomainType_ field.
+
+```go
+
+type IdentifierF func() []byte // <-- new func type
+type DomainTypeF func() (types.DomainType,error) // <-- new func type
```
GalRogozinski, line 200:
```suggestion
type DomainTypeF func() (types.DomainType, error) // <-- new func type
```

not a blocker for merging

sips/fork_support.md
```diff
@@ -161,7 +160,7 @@ func (networkID NetworkID) ForkAtEpoch(epoch phase0.Epoch) (*ForkData, error) {
 
 	// If empty, raise error
 	if len(forks) == 0 {
-		return nil, errors.New("GetCurrentFork: fork list by GetForksData is empty.")
+		return nil, errors.New("GetCurrentFork: fork list by GetForksData is empty. Unknown Network")
```
alonmuroch, line 163: let's keep error format to what we use currently. "GetCurrentFork:..." is not how we write errors

alonmuroch (changes requested):

sips/fork_support.md
```diff
@@ -20,56 +20,254 @@ Client implementation should calculate when a new duty starts to decide if it st
...
+type BaseRunner struct {
+	State          *State
+	Share          *types.Share
+	QBFTController *qbft.Controller
+	BeaconNetwork  types.BeaconNetwork
+	BeaconRoleType types.BeaconRole
+	highestDecidedSlot spec.Slot
+}
+
+func (b *BaseRunner) GetIdentifierF() func() []byte {
```
alonmuroch, line 254: why does it return a func?

sips/fork_support.md
```diff
@@ -20,56 +20,254 @@ Client implementation should calculate when a new duty starts to decide if it st
...
+	return func() []byte {
+		currentEpoch := b.BeaconNetwork.EstimatedCurrentEpoch()
+		domainType := b.Share.NetworkID.ForkAtEpoch(currentEpoch)
+		identifier := types.NewMsgID(domainType,b.Share.ValidatorPubKey[:],b.BeaconRoleType)
+		return identifier[:]
+	}
+}
+
+func (b *BaseRunner) GetDomainTypeF() func() (types.DomainType, error) {
+	return func() (types.DomainType, error) {
```
alonmuroch, line 264: why does it return a func?

alonmuroch (changes requested):

sips/fork_support.md
GalRogozinski: Because the identifier is a function of the current epoch

sips/fork_support.md
GalRogozinski:
```suggestion
		return nil, errors.New("Fork list by GetForksData is empty. Unknown Network")
```

sips/fork_support.md
```diff
@@ -20,56 +20,254 @@ Client implementation should calculate when a new duty starts to decide if it st
...
+}
+
+func (c *Controller) StartNewInstance(height Height, value []byte) error {
+	// ...
+
+	domainType, err := c.DomainTypeF()
+	if err != nil {
+		return Errors.Wrap(err,"Coudld not get current DomainType.")
+	}
+	c.config.SetSignatureDomainType(domainType) // <-- updates its config using its new DomainF
```
alonmuroch, line 219: config is being injected into the controller and provides data/ functions. The controller shouldn't be "setting" stuff in the config

sips/fork_support.md
alonmuroch: I dislike the use of functions in this case, it makes it harder to read and test. 

Why not simply have a setter function for the runner/ controller/ etc.

sips/fork_support.md
alonmuroch: see my comments above

alonmuroch (approved):

GalRogozinski (approved):
Latest changes look good 👍

sips/fork_support.md
```diff
@@ -20,56 +20,231 @@ Client implementation should calculate when a new duty starts to decide if it st
...
+
+Regarding the indirect usage of Identifiers:
+- `Instance`:
+  - Must keep its identifier to send properly formed messages.
+- `Controller`:
+  - Should have its _Identifier_ and _DomainType_ updated when a new fork occurs. For that, it will have setter functions for both fields. These functions will be used by the `Runner` to update the controller.
+
+```go
+
+type Controller struct {
```
alonmuroch, line 198: please update to current controller struct.. this seems old

GalRogozinski (changes requested):
From @alonmuroch:
1. Remove all the duplicate source of truths of Domain. Attempt to use `Share` as the single source of truth.
2. Setting domain is "implementation detail"

sips/fork_support.md
MatheusFranco99: Indeed. Updating, thanks

sips/fork_support.md
```diff
@@ -20,56 +20,212 @@ Client implementation should calculate when a new duty starts to decide if it st
...
+func (c *Controller) SetIdentifier(identifier []byte) {
+	c.Identifier = identifier
+}
+
+func (c *Controller) SetDomainType(domainType types.DomainType) {
+	c.Domain = domainType
+}
+```
+
+For the SSV team implementing the Fork update, we advise the following runner's function to get the appropriate domain type for its controller.
```
GalRogozinski, line 220:
```suggestion
The following runner's function gets the appropriate domain type for its controller.
```

No need to mention the SSV team when we define the protocol

sips/fork_support.md
```diff
@@ -20,56 +20,212 @@ Client implementation should calculate when a new duty starts to decide if it st
...
+- `Controller`:
+  - Should have its _Identifier_ and _DomainType_ updated when a new fork occurs. For that, it will have setter functions for both fields. These functions will be used by the `Runner` to update the controller.
+
+```go
+
+type Controller struct {
+	Identifier []byte
+	Height     Height
+	StoredInstances InstanceContainer
+	Domain          types.DomainType
```
GalRogozinski, line 206: Refresh my memory.
We said we can get rid of Domain here and use it just in config, no?

GalRogozinski (changes requested):

sips/fork_support.md
```diff
@@ -91,14 +91,14 @@ Along with the `DomainType`, we include a `NetworkID` type and a new `ForkData`
...
-	JatoNetworkID    = NetworkID(0x3)
-	JatoV2NetworkID    = NetworkID(0x4)
+type NetworkID [1]byte
+
+var (
+	MainnetNetworkID = NetworkID{0x0}
+	PrimusNetworkID  = NetworkID{0x1}
+	ShifuNetworkID   = NetworkID{0x2}
+	JatoNetworkID    = NetworkID{0x3}
+	JatoV2NetworkID  = NetworkID{0x4}
```
GalRogozinski, line 101: I have no issues with this change.
I am just wondering why you decided to do it?  Seems like an implementation detail.

MatheusFranco99: It's an implementation detail indeed. I made the change due to marshaling problems.

sips/fork_support.md
MatheusFranco99: We can. I'll update it

sips/fork_support.md
```diff
@@ -20,56 +20,209 @@ Client implementation should calculate when a new duty starts to decide if it st
...
+	V3Testnet      = DomainType{0x0, 0x0, 0x3, 0x1}
+)
+```
+The _DomainType_ major importance is the fact that **it's mixed in the signature** such that messages on different domains will have different signatures.
+
+The node takes one of the above _DomainType_ values and **holds it until termination**.
+
+In the _DomainType_, the third byte corresponds to the **network** (Mainnet, Primus, etc.), and the fourth byte corresponds to the **fork or protocol version** for this network.
+
+In the actual implementation, _DomainType_ appears in the following structures:
```
alonmuroch, line 43: actual "spec" not implementation

sips/fork_support.md
```diff
@@ -20,56 +20,209 @@ Client implementation should calculate when a new duty starts to decide if it st
...
+	}
+	ret := sha256.Sum256(byts)
+	return ret[:], nil
 }
+```
+
+### Structures access to types
+
+- `Share`:
+  - It's more appropriate for the structure to hold the information on the network that it's in, instead of the domain type. Thus, _Share_ will contain a `NetworkID` field instead of a `DomainType` field.
```
alonmuroch, line 190: didn't we say we are reverting this change?

GalRogozinski: If we don't do this change implementation team will be having issue imo

sips/fork_support.md
```diff
@@ -20,56 +20,209 @@ Client implementation should calculate when a new duty starts to decide if it st
...
+
+func (c *Controller) SetIdentifier(identifier []byte) {
+	c.Identifier = identifier
+}
 ```
 
+The following runner's function gets the appropriate domain type for its controller.
+```go
+// GetDomainTypeAtSlot returns the domain type for a given slot using the its BeaconNetwork and Share's NetworkID
+func (b *BaseRunner) GetDomainTypeAtSlot(slot phase0.Slot) (types.DomainType, error) {
```
GalRogozinski, line 220: make this a regular function and not a method please so we don't have to necessarily add it to spec

sips/fork_support.md
GalRogozinski: @MatheusFranco99 
After talking to Alon we will keep DomainType in share in the SIP.

Since currently Share is held by the `Validator` it may seem problematic.

However since there are no tests for `Validator` we can either easily change it in spec code or even maybe remove it (if it isn't tested maybe no need for it).

This will allow engineering to be able to work with the SIP because they won't be constrained on where `share` is placed

sips/fork_support.md
```diff
@@ -20,56 +20,204 @@ Client implementation should calculate when a new duty starts to decide if it st
...
+	GenesisMainnet = DomainType{0x0, 0x0, 0x0, 0x0}
+	PrimusTestnet  = DomainType{0x0, 0x0, 0x1, 0x0}
+	ShifuTestnet   = DomainType{0x0, 0x0, 0x2, 0x0}
+	ShifuV2Testnet = DomainType{0x0, 0x0, 0x2, 0x1}
+	V3Testnet      = DomainType{0x0, 0x0, 0x3, 0x1}
+)
+```
+The _DomainType_ major importance is the fact that **it's mixed in the signature** such that messages on different domains will have different signatures.
+
+The node takes one of the above _DomainType_ values and **holds it until termination**.
```
GalRogozinski, line 39: I think now this sentence is confusing

sips/fork_support.md
```diff
@@ -20,56 +20,204 @@ Client implementation should calculate when a new duty starts to decide if it st
...
+	...
+	signedMsg.GetSignature().VerifyByOperators(signedMsg, b.Share.DomainType, types.PartialSignatureType, b.Share.Committee)
+	...
+}
+```
+
+- Indirect usage: an _Instance_ includes its _State.ID_ in messages. The _State.ID_ is initialized as the controller's _Identifier_ which contains the _DomainType_. The controller's identifier is fixed. It's used as:
+  - an argument on instance creation.
+  - to broadcast decided messages.
+  - in base validation, to verify if the message was sent to the correct controller.
```
GalRogozinski, line 83: let's remove implementation details.

GalRogozinski (changes requested):

sips/fork_support.md
MatheusFranco99: Do you mean the "The node takes one of the above _DomainType_ ..." sentence? Notice that it describes the previous state

sips/fork_support.md
MatheusFranco99: Ok, removing

sips/fork_support.md
GalRogozinski: ahh ok, my bad

GalRogozinski (approved):

alonmuroch (approved):

PR #35, opened 2023-10-09
RSA SIP

MatheusFranco99:


sips/rsa_network_authentication.md
```diff
@@ -0,0 +1,36 @@
...
+4. Cryptographic Agility: While RSA is currently a widely accepted encryption and digital signature algorithm, the field of cryptography is constantly evolving. It's important to design the system with cryptographic agility in mind, allowing for the future transition to more secure algorithms if necessary. This flexibility ensures that the network can adapt to emerging security threats and maintain a robust security posture over time. It's worth noting that RSA, despite its prevalence, is among the most targeted cryptographic schemes. For instance, elliptic curve cryptography is often recommended due to its enhanced security features. However, even these might fall by the wayside with the impending rise of quantum computers, compelling a transition to advanced alternatives like Dilithium or Falcon.
+5. Compliance and Standards: Ensure that the implementation adheres to industry standards and best practices, such as those outlined by NIST and other relevant regulatory bodies. Compliance with established standards helps validate the security of the RSA implementation and provides assurance against common vulnerabilities and exploits.
```
GalRogozinski, line 32: 1 & 5. This in my opinion can be moved to the specification section. We just need to say that we're going to be using key length according to NIST specifications.
2. I don't see this as a security concern for us but as a drawback of RSA compared to modern EC schemes that have smaller keys. Maybe you can discuss this in "Rationale", "Drawbacks" or "Discussed Altrernatives" (your choice).
3. This should go into specification. No need to discuss OAEP. The point of this document is to explain how we do authentication with RSA and why. No need to go wild here :-)
4.I don't see why this point is relevant to be here. We are not doing anything to make our system cryptographically agile... in my opinion we don't need this point at all.

GalRogozinski (changes requested):

sips/rsa_network_authentication.md
MatheusFranco99: 1,5,2,3 -> Sure, changed :)
4 -> The point I tried to make was to highlight that we may need to change the scheme in the future and, thus, the implementation should offer some abstraction of these layers since this is likely to change in some years. Do you think it should be removed? (passing it to specification btw)

sips/rsa_network_authentication.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Specification
+
+At present, the operator employs their BLS private key share for verification. However, the operator also possesses an RSA key, which would replace the BLS key for signing messages. All other nodes store both BLS and RSA public keys of every node, eliminating the need for a key exchange protocol.
```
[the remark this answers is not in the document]
MatheusFranco99: We will not use libp2p to authenticate.
Moving it 👍

GalRogozinski (commented):
another thing that we need to specify is when we we still use BLS. We said that before extrapolating the signature we should check for BLS validity to thrawt attacks

GalRogozinski (changes requested):
The most important part is the specification. We need to write that we have a check that the RSA public key is known to the contract (you implied it in other sections but it should be specified).

Also write down which specific padding scheme should be used (PSS) without "or".

When I read the Specification section it sounds more like some preliminary intro to RSA and crypto. This is not the point of this section. In general I don't think this should be in the SIP.

It should be concise and give clear instructions to the implementor. Usually with pseudo-code but not necessarily.
For example in one line you can write that we use a 2048 bits key with PSS. In another line write that we have to be compliant with standards (and link to NIST) and that would be enough.

sips/rsa_message_authentication.md
```diff
@@ -0,0 +1,76 @@
...
+
+
+## Appendix
+
+### A1 - Security Considerations
+
+The following security and design considerations should be thoroughly addressed:
+
+1. Key Length and Security: The security of RSA encryption is highly dependent on the length of the key used. For instance, shorter keys are more vulnerable to brute-force attacks. The [US National Institute of Standard and Technology (NIST)](https://www.nist.gov) approves a minimum of 2048-bit RSA keys. Check the first table of section 1.5 of their [Security Policy](https://csrc.nist.gov/CSRC/media/projects/cryptographic-module-validation-program/documents/security-policies/140sp4172.pdf), released in 2023 July, for this reference.
+2. Padding Schemes: RSA signatures require the use of padding schemes to ensure security. Poorly implemented or outdated padding schemes can expose the system to padding oracle attacks, where an attacker can gain unauthorized access to encrypted data. Employing secure padding schemes, such as PKCS#1 v1.5 or PSS, is essential to prevent these vulnerabilities.
```
GalRogozinski, line 64: write down which exact specific padding scheme is used

GalRogozinski (changes requested):

GalRogozinski:
Great!
I am waiting with the merge to give pressure to check the PSS recommendation!

GalRogozinski (approved):

PR #37, opened 2023-12-25
New-qbft-timeout

GalRogozinski:
# Description

New timeout rules.

From the slot start time after `baseDuration` delay wait 2 seconds before emitting a `RoundChange` message until the 8th round (inclusive). Afterwards, wait 2 minutes.

### DIFF
--- a/sips/dynamic_qbft_timeout.md
+++ b/sips/dynamic_qbft_timeout.md
@@ -4,7 +4,7 @@
 
 
 **Summary**  
-Changes to the constant timeouts we defined in SIP#6 that should improve performance. 
+Changes to the constant timeouts we defined in [SIP#6](https://github.com/bloxapp/SIPs/blob/main/sips/constant_qbft_timeout.md) that should improve performance. 
 
 **Rational**  
 Currently for all duties, we have a `quickTimeout` of 2 seconds before sending a `RoundChange` message and a `slowTimeout` of 2 minutes starting from the 8th round.The timer starts when the QBFT instance starts. This approach has 2 issues:
@@ -27,7 +27,9 @@ The Beacon chain has some timing assumptions regarding duty execution for maximi
 
         
 
-**Specification** 
+**Specification**
+
+From the slot start time after `baseDuration` delay wait 2 seconds before emitting a `RoundChange` message until the 8th round (inclusive). Afterwards, wait 2 minutes.
 
 ```go
 var (
@@ -36,49 +38,48 @@ var (
     slowTimeout           = 2 * time.Minute
 )
 
-
-// Timeout returns the number of seconds until next timeout for a given round.
-// When we have the proposal duty we immediately call it.
-func Timeout(r Round) time.Duration {
-    if r <= quickTimeoutThreshold {
-        return quickTimeout
-    }
-    return slowTimeout
-}
-
+// RoundTimeout returns the time in which we should send a RC message
 // Called for all beacon duties other than proposals
-func RoundTimeout(r Round, baseDuration Duration) Time {
-    if r == Round.FirstRound {
-        return baseDuration + quickTimeout
-    } else {
-        return Timeout(r)
-    }
+func RoundTimeout(r Round, dutyStartTime Time, baseDuration Duration) Time {
+   // Calculate additional timeout based on round
+	var additionalTimeout time.Duration
+	if round <= quickTimeoutThreshold {
+		additionalTimeout = time.Duration(int(round)) * slowTimeout
+	} else {
+		quickPortion := time.Duration(quickTimeoutThreshold) * quickTimeout
+		slowPortion := time.Duration(int(round-quickTimeoutThreshold)) * slowTimeout
+		additionalTimeout = quickPortion + slowPortion
+	}
+
+	// Combine base duration and additional timeout
+	timeoutDuration := baseDuration + additionalTimeout
+
+	// Get the start time of the duty
+	dutyStartTime := t.beaconNetwork.GetSlotStartTime(phase0.Slot(height))
+
+	// Calculate the time until the duty should start plus the timeout duration
+	return time.Until(dutyStartTime.Add(timeoutDuration))
 }
 ```
 
 *Proposals*
 
-No changes from SIP#6.
-
-```go
-func ProposalTimeout(r Round)
-    return Timeout(r)
-```
+No changes from [SIP#6](https://github.com/bloxapp/SIPs/blob/main/sips/constant_qbft_timeout.md). Doesn't rely on the slot/duty start time.
 
 *Attestations/Sync Committees*
 
-Once the slot starts, wait 4 seconds for a block to be produced then start the timer according to the same rules as defined in SIP#6.
+Once the slot starts, wait 4 seconds for a block to be produced then start the timer according to the new rules.
 
 ```go
-func AttestationOrSyncCommitteeTimeout(r Round)
-    return RoundTimeout(r, 4)
+func AttestationOrSyncCommitteeTimeout(r Round, dutyStartTime Time) Time
+    return RoundTimeout(r, dutyStarTime, 4)
 ```
 
 *Aggregator/Contribution*
 
-Once the slot starts, wait 8 seconds for a block to be produced then start the timer according to the same rules as defined in SIP#6.
+Once the slot starts, wait 8 seconds for a block to be produced then start the timer according to the new rules
 
 ```go
-func AggregationOrContribuitionTimeout(r Round)
-    return RoundTimeout(r, 4)
+func AggregationOrContribuitionTimeout(r Round, dutyStartTime Time) Time
+    return RoundTimeout(r, dutyStartTime 8)
 ```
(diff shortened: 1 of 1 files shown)


MatheusFranco99 (approved):
Good Work

sips/dynamic_qbft_timeout.md
```diff
@@ -0,0 +1,85 @@
...
+
+        
+
+**Specification**
+
+From the slot start time after `baseDuration` delay wait 2 seconds before emitting a `RoundChange` message until the 8th round (inclusive). Afterwards, wait 2 minutes.
+
+```go
+var (
+    quickTimeoutThreshold = Round(8)
```
GalRogozinski, line 36: I am actually not sure why this is so high... 
On Round 6 the slot already ended

MatheusFranco99 (dismissed):
Well changed

sips/dynamic_qbft_timeout.md
```diff
@@ -0,0 +1,85 @@
...
+	dutyStartTime := t.beaconNetwork.GetSlotStartTime(phase0.Slot(height))
+
+	// Calculate the time until the duty should start plus the timeout duration
+	return time.Until(dutyStartTime.Add(timeoutDuration))
+}
+```
+
+*Proposals*
+
+No changes from [SIP#6](https://github.com/bloxapp/SIPs/blob/main/sips/constant_qbft_timeout.md). Doesn't rely on the slot/duty start time.
```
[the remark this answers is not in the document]
y0sher: sounds to me like we do need to do something with proposals. , a dynamic timeout will solve an issue where some nodes might receive the block at different times (different beacon nodes) and then start and end the rounds on different times.

y0sher (changes requested):
generally looks good, I think proposals need a little bit more thinking, I remember @olegshmuelov had something to say about it.

sips/dynamic_qbft_timeout.md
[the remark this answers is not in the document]
GalRogozinski: @moshe-blox 

1. We start the timer at the duty start time
2. We need to configure the timeout parameters to something reasonable
3. If you are not the leader of the first round then prepare a full block for the next rounds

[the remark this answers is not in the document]
GalRogozinski: @moshe-blox 
The issue is for selecting reasonable timeout parameters, if the leader of the first round is bad and timeout is too slow we risk proposing the block late

GalRogozinski:
@y0sher 

We will take care of proposals in a different place

y0sher:
> @y0sher 
> 
> We will take care of proposals in a different place

Please lets mention in this SIP that it doesn't take into consideration proposals

sips/dynamic_qbft_timeout.md
GalRogozinski: 6 is a more logical choice for more reasons even:
1. For 4 nodes committee we would have 2 nodes doing quick-timeouts twice
2. For 7 nodes committee, 6 nodes would do quick timeouts and the last node a slow one. But we managed to circle all the nodes.
3. For 13 nodes committee, 7 nodes would be more than half the committee...

Also later, not in this SIP, we may decide that there is no point to keep on doing consensus after the end of the slot, so this would be a preparation...

GalRogozinski:
@y0sher 

The current reference is not enough?
https://github.com/bloxapp/SIPs/pull/37/files#diff-cd9cb4d346903b051ff59aa32693b26216634cfcbface288021be06b912f2e82R66-R68

sips/dynamic_qbft_timeout.md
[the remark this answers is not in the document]
GalRogozinski: proposals won't be considered in this SIP

y0sher:
> @y0sher
> 
> The current reference is not enough? https://github.com/bloxapp/SIPs/pull/37/files#diff-cd9cb4d346903b051ff59aa32693b26216634cfcbface288021be06b912f2e82R66-R68

IMO No. should be part of the SIP itself.

sips/dynamic_qbft_timeout.md
moshe-blox: @GalRogozinski
agreed, if this change would be a fork, then i'd suggest to also reduce `CutoffRound` to 1 slot for proposal and sync committee duties to reduce RC spam (afaik, they can't be retried after the slot has ended)

moshe-blox (approved):

sips/dynamic_qbft_timeout.md
```diff
@@ -0,0 +1,90 @@
+| Author         | Title                 | Category | Status   |
+| -------------- | --------------------- | -------- | -------- |
+| Gal Rogozinski |  QBFT timeout | Core     | approved |
+
+
+**Summary**  
+Changes to the constant timeouts we defined in [SIP#6](https://github.com/bloxapp/SIPs/blob/main/sips/constant_qbft_timeout.md) that should improve performance. 
+
+**Rational**  
+Currently for all duties, we have a `quickTimeout` of 2 seconds before sending a `RoundChange` message and a `slowTimeout` of 2 minutes starting from the 8th round.The timer starts when the QBFT instance starts. This approach has 2 issues:
```
iurii-ssv, line 10: Could you clarify btw why there is a need for `slowTimeout` ? Is it just purely for qbft or do we have anything in SSV node that would make use of such long timeout (I think most/all duties must finish execution much faster or just will be missed)

nkryuchkov: @iurii-ssv It was introduced in https://github.com/ssvlabs/SIPs/pull/22/files, attester duties may be submitted within 32 slots/6.4m

sips/dynamic_qbft_timeout.md
```diff
@@ -0,0 +1,90 @@
...
+**Specification**
+
+From the slot start time after `baseDuration` delay wait 2 seconds before emitting a `RoundChange` message until the 6th round (inclusive). Afterwards, wait 2 minutes.
+
+```go
+var (
+	// quickTimeoutThreshold is the round after which the timeout duration increases to 2 minutes
+	quickTimeoutThreshold = Round(6)
+	// quickTimeout is the timeout in seconds for the first 6 rounds
+	quickTimeout int64 = 2 // 2 seconds
```
nkryuchkov, line 39: IMO `time.Duration` is more readable

```suggestion
	quickTimeout time.Duration = 2 * time.Second
```

sips/dynamic_qbft_timeout.md
GalRogozinski: The idea is simply wait for a very long time if after 8 rounds you didn't reach consensus. This is because we are probably in a very high latency environment

sips/dynamic_qbft_timeout.md
GalRogozinski: I was trying to move away from using go libs in SIPs
The idea is that the spec shouldn't be for helping implement a node in any language

Also for future SIPs I want to start using python like pseudo-code

GalRogozinski:
Hey @iurii-ssv and cc @y0sher 

So this SIP was supposed to simply document the changes that the SSV team did to the protocol without talking with the spec team (well I was on vacation). It didn't get merged due to shifting priorities.

Since it wasn't merged, we can do proposal solutions on this SIP maybe.
But currently, I prefer to merge it as it is along with the accompanying spec PR https://github.com/ssvlabs/ssv-spec/pull/352. This is because it is hanging for a long time. 

Btw, there is a PR for your idea of doing an asynchronous call https://github.com/ssvlabs/ssv-spec/pull/371. It wasn't merged due to shifting priorities.

Note that RC may degrade performance temporarily if not done in a coordinated fork

sips/dynamic_qbft_timeout.md
iurii-ssv: > @iurii-ssv It was introduced in https://github.com/ssvlabs/SIPs/pull/22/files, attester duties may be submitted within 32 slots/6.4m

right, 

at any rate, I always thought the **spec** should define an interface (not an implementation) everybody wants to conform to and implement it as they see fit for their use-case/application, meaning:
- QBFT shouldn't have to know anything about duty(s), it should be the other way around - duty configures QBFT Instance to work as it needs to
- different duty types should have different `total round count`, `round 1 timeout`, `round 2 timeout`, ... (seems like these values should not be constants, but rather vars - so it doesn't send the wrong message that "these values are uniform for everyone")

It's not really an issue to worry about, just trying to get some consistency (and knowledge) if possible.

sips/dynamic_qbft_timeout.md
iurii-ssv: Just my 2 cents,

I would say while we still use Golang might as well do it as Golang prefers to do it (as Nikita mentioned), but the idea to use pseudo-code is the right one ofc

the even more important thing though is that spec "provides interface and not implementation" - I elaborate a bit on it [here](https://github.com/ssvlabs/SIPs/pull/37#issuecomment-2445303530)

GalRogozinski: > as long as we keep away from using advanced Python-specific syntax
That's why I said "python like pseudo code" :-) 

> spec "provides interface and not implementation" 
Currently spec has too many implementation details including full fledged consensus impl.
This is due to legacy reasons and how the project was inherited. We are planning a Markdown spec.
All implementations that we will have will just be used for creating tests.

PR #40, opened 2024-01-30
Verifiable Consensus Data

GalRogozinski:


sips/verifiable_consensus_data.md
```diff
@@ -0,0 +1,235 @@
...
+	validatorIndex phase0.ValidatorIndex,
+	sharePublicKey []byte,
+    // obtained from the operator's beacon node
+    attestationDataByts []byte,
+) qbft.ProposedValueCheckF {
+	return func(data []byte) error {
+        // addition
+        if bytes.Equal(attestationDataByts, data) {
+            return nil
+        }
```
MatheusFranco99, line 112: This is a `nil` or empty check right?

sips/verifiable_consensus_data.md
```diff
@@ -0,0 +1,235 @@
...
+func verifyData(expectedRoot phase0.Root, proof phase0.SignedBeaconBlockHeader) error {
+    beaconRoot := proof.Message.HashRoot()
+    if !bytes.Equal(expectedRoot, beaconRoot) {
+        return errors.New("unexpected header root")
+    }
+    blockEpoch := getBeaconNode().getEpochBySlot(proof.Message.slot)
+    domain := GetBeaconNode().DomainData(blockEpoch, spec.DomainProposer)
+    root := spec.ComputeEthSigningRoot(beaconRoot, domain)
+    validator := GetBeaconNode.GetValidator("HEAD", proof.Message.ProposerIndex)
+    return bls.Verify(sig, validator.PubKey, root)
```
MatheusFranco99, line 96: With this check, we're sure that the block proposed as head exists. But we, still, can't be sure if a malicious operator is sending us an old block to attest, correct?
Perhaps, if we had some state that persisted through all duties, we could do a lower limit check for the slot/epoch. This is some future discussion but what do you think?

MatheusFranco99 (dismissed):
Awesome!!

sips/verifiable_consensus_data.md
GalRogozinski: We can compare against the block we got from our node simply without persisting state.
But I see it as a future improvement outside this SIP

sips/verifiable_consensus_data.md
GalRogozinski: if the attestation data we got from our own node is the same as the leader's (it should happen if the nodes are synced to the same head) then no need to perform any more validation and we stop

This assumes the operator trusts his beacon node (most of the time they run their own I think).

If the `data` is `nil` it should fail I think when we do `deocode` or `validate`

GalRogozinski:
I need to add a check that the beaconHeaderProof is assigned by a validator that was assigned the duty

alonmuroch:
Why can't a local node request the block from the local beacon node and see it exists?

GalRogozinski (commented):
> Why can't a local node request the block from the local beacon node and see it exists? 

it can but the block may be missing there. We first check the local node and then the proof

sips/verifiable_consensus_data.md
```diff
@@ -0,0 +1,284 @@
...
+    if attestationData.Target.Epoch == currentEpoch {
+        if attestationData.Source.Root != checkpoints.data.currentJustifiedRoot {
+            return Errors.New("source is not expected currently justified root")
+        }
+    } else if attestationData.Source.Root!= checkpoints.data.previousJustifiedRoot {
+        return Errors.New("source is not expected previously justified root")
+    }
+
+    return nil
+}
```
GalRogozinski, line 179: This checks exists in [phase0](https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#attestations).

In [deneb](https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/beacon-chain.md#modified-get_attestation_participation_flag_indices) it is hidden inside the `participation flags` that check eligibility for rewards. There is still an `assert` statement for this check.

Issue #39, opened 2024-01-29
Boole Fork Wish List


GalRogozinski:
## Description

An ever-updated list of changes we want that need a fork.

### In
- [ ] #43 (performance & scalability)
- [ ] #61  (performance & scalability)
- [ ] #65 (security)
### Out
- [ ] #63 (performance)
- [ ] #67 (security) (because it isn't a forking change)
- [ ] [If not QBFT leader, do not wait for start value](https://github.com/ssvlabs/ssv-spec/pull/371) - less important because currently no issues with proposals.
  **Prerequisites:**
  - [ ] #37 
  - [ ] https://github.com/ssvlabs/ssv-spec/pull/352
- [ ] https://github.com/ssvlabs/ssv-spec/issues/532
- [ ] #58 - probably lots of work
- [ ] https://github.com/ssvlabs/ssv-spec/issues/373 - No issues with proposals
- [ ] #59 
- [ ] https://github.com/ssvlabs/ssv-spec/pull/355

GalRogozinski:
This list was continued in https://github.com/ssvlabs/ssv-spec/issues/368

GalRogozinski:
@moshe-blox 
you mean https://github.com/ssvlabs/ssv-spec/pull/371? we can do it

MatheusFranco99:
[Fork: QBFT Identifier with fixed length -> msg encoding changes](https://github.com/ssvlabs/ssv-spec/pull/536)

MatheusFranco99:
Remove repeated Signer field in `spec.types.PartialSignatureMessage` (already exists in `spec.types.PartialSignatureMessages`).

GalRogozinski:
Moved to #77

PR #42, opened 2024-03-06
Committee-based consensus

MatheusFranco99:
This SIP proposes aggregating `Attestation` and `Sync Committee` duties on a single duty. This will reduce the number of messages exchanged and processing costs.


sips/cluster_consensus.md
```diff
@@ -0,0 +1,127 @@
+|     Author     |           Title            |  Category  |       Status        |    Date    |
+| -------------- | -------------------------- | ---------- | ------------------- | ---------- |
+| Matheus Franco | Cluster consensus          | Core       | open-for-discussion | 2024-03-05 |
+
+## Summary
+
+Aggregate `Attestation` and `Sync Committee` duties based on the cluster of operators and the duties' slot. Also, merge the post-consensus partial signature messages for such duties into a single message.
```
GalRogozinski, line 7: I don't see in the SIP nothing about merging post consensus?

sips/cluster_consensus.md
```diff
@@ -0,0 +1,127 @@
...
+	subgraph MessageID
+		Domain
+		OperatorCluster
+		Role
+	end
+```
+
+### QBFT Controller
+
+The current `QBFTController` structure allows only one consensus instance at a time. This must change, extending the `QBFTController` into a router of messages for the different instances.
```
GalRogozinski, line 90: In my opinion we can simply drop `QbftController` from spec
It used to be there when a runner ran several instances at the same time...

We don't do this anymore.
Even if it is needed by the implementation people I don't see why spec needs this

sips/cluster_consensus.md
```diff
@@ -0,0 +1,127 @@
...
+```go
+func (r *DutyRunner) execute() {
+	r.QBFTController.StartConsensus(r.committee, r.duty, r)
+}
+
+func (c *QBFTController) StartConsensus(committee []types.Operator, duty types.Duty, observer Observer) {
+	instance := c.StartInstance(committee, duty)
+	instance.registerObserver(observer)
+}
+```
```
GalRogozinski, line 107: > Since the `DutyRunner` will not have its specific `QBFTController`

This wording is confusing since you later show that the runner has a direct reference to `QBFTController`

Also we shouldn't get into implementation details with the observer unless absolutely necessary for tests

sips/cluster_consensus.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Drawbacks
+
+- Several duties will depend on the same consensus execution. Thus, its failure will imply many attestation misses. Nonetheless, this will also make the operator more scalable due to the overall reduction in exchanged messages and processing costs.
+
+## Extra improvements
+
+- If multiple partial signatures contained in a merged message refer to the same root (i.e. validators in the same Ethereum committee), the signatures can be verified using batch verification.
```
GalRogozinski, line 115: this is just the reduce signatures idea?
If so does it belong here?

sips/cluster_consensus.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Message Validation
+
+This duties transformation requires similar changes in message validation, namely:
+- Different consensus executions are tagged by the `MessageID`. This change would be propagated with no further issues. However, the `MessageID` is used to get the validator's public key and the duty's role which are used as an ID to store the consensus state. This must be changed to use the operators' committee and the duty's role, or even simply the `MessageID`.
+- Message validation limits the number of attestation duties per validator by using the validator's public key contained in the `MessageID`. This is no longer possible. A new limitation can be accomplished by checking the number of validators a cluster of operators is assigned to. If this number is less than 32 (the number of slots in an epoch), then we can limit the number of attestation duties of such cluster per epoch. The only exception would be if such a cluster is assigned to a sync committee duty (considering that we will indeed merge attestations and sync committee duties altogether in the same consensus execution).
```
GalRogozinski, line 121: you also update message validation in knowledge base
should be on a seperate branch...
Also google sheet should be versioned

sips/cluster_consensus.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Open questions
+
+- What should be the maximum number of signatures a post-consensus message can contain? The trade-off here refers to reducing the number of exchanged messages versus reducing the impact of a DoS buffer attack attempt.
```
GalRogozinski, line 125: didn't understand this and the relation to the SIP

GalRogozinski (changes requested):

GalRogozinski:
I think this requires reduce signatures with a fork to totally take out BLS out of consensus?

sips/cluster_consensus.md
MatheusFranco99: Added more detail

sips/cluster_consensus.md
MatheusFranco99: Specifically for this change, I think we should keep it to handle the several operator's QBFT instances running in parallel.

sips/cluster_consensus.md
MatheusFranco99: I see why this is confusing. Sorry for that.

What I meant is that the `DutyRunner` used to have a `QBFTController` of its own. Now, the operator has the `QBFTController` and the `DutyRunner` will hold a reference to it.

I tried not to add a lot of code. I thought that would help to show how the observer pattern could be used.

sips/cluster_consensus.md
MatheusFranco99: Sure. Will do it after all the changes here are agreed on.

sips/cluster_consensus.md
MatheusFranco99: In this SIP, I suggest merging the post-consensus signatures of the aggregated duties into a single message. However, I also say that the number of signatures in a message must be bounded. Here, I leave this issue to be opened for discussion.

MatheusFranco99:
@GalRogozinski For this change, we can still use BLS in consensus.

GalRogozinski:
@MatheusFranco99 when you are doing operator based consensus, which validator do you choose?

What happens when validators are added or removed? You didn't spec this in the SIP...

You can find logic to solve this, but I think it makes more sense to simply switch to operator's keys... Because we should do this anyhow

MatheusFranco99:
@GalRogozinski 
I think you don't need to choose any validator for that. As I see it, if you are operator A and validator X with cluster (A, B, C, D) has a duty, then you would start a consensus instance for the cluster (A, B, C, D) for the duty's slot. The consensus data can either be an `AttestationData` with any `CommitteeIndex`

```go
type AttestationData struct {
	Slot             Slot
	Index           CommitteeIndex // Default value
	BeaconBlockRoot Root 
	Source         *Checkpoint
	Target          *Checkpoint
}
```
or a custom structure like:
```go
struct {
	BeaconBlockRoot Root 
	Source         *Checkpoint
	Target          *Checkpoint
}
```

Another validator Y with the same cluster would use the same consensus instance and get the useful decided information about the `Root` and `Checkpoints` to build the object to be signed.

MatheusFranco99:
@GalRogozinski
Do you mean aggregating the validator registration and exit duties? Or how added/removed validators change this aggregation?
For the second question, removed validators shouldn't have the duty event and won't be using the decided data by its cluster's consensus, while added validators would use this data.

GalRogozinski:
You have validators X and Y and Z on a single cluster. You need logic to
decide which validator to use.

Upon validator entry/exit the chosen validator should change.

So you need to spec a function that chooses a deterministic validator based
on some logic each time consensus starts

MatheusFranco99:
@GalRogozinski What do you mean by choosing a validator?

MatheusFranco99:
If operators used BLS for consensus, they would need to choose a validator to know with which BLS key they should sign messages. On discussion, we concluded that this approach would require some agreed-upon deterministic selection. This would not be easy with the current design (due to the decoupling of consensus and duty execution logic).

However, there's a cryptography change going on by which operators should use their network keys to validate consensus messages. Relying on that would make things much simpler. Therefore, we prefer to make this change a pre-requisite for this SIP.

moshe-blox:
Stuff to think about:
- Subnet calculation by cluster identity? And if so — we must refactor subnet scoring to tune expected message rate

moshe-blox:
Just noting that this SIP should have the representation of the [BeaconState](https://github.com/bloxapp/knowledge-base-internal/discussions/13#BeaconState) (or however you decide to call it) and a specification of how to construct it.

MatheusFranco99:
@moshe-blox Indeed. Will add it later on. Thanks!

MatheusFranco99:
I like the design @moshe-blox .

This cluster-based consensus PR will already be huge anyway. I think this design produces an even bigger change in terms of code and tests, but I think it may be worth the cost.

GalRogozinski:
New containers and interfaces to be added:

```
// Cluster is a cluster of a unique set of operators that run the same validator set
type Cluster struct {
	ConsensusRunner ConsensusRunner
	SSVRunners      []SSVRunner
	Network         Network
	Beacon          BeaconNode
	Shares           *[]types.Share
	Signer          types.SSVSigner
}

type ConsensusRunner interface {
	Getters

	GetSSVSigner() types.SSVSigner
	StartNewInstance(role types.BeaconRole, slot spec.Slot) error
	ProcessConsensus(msg *qbft.SignedMessage) error
	HasRunningInstance() bool
}

type SSVRunner interface {
	Getters

	GetBeaconSigner() types.BeaconSigner
	StartNewDuty(duty *types.Duty) error
	ProcessPreConsensus(signedMsg *types.SignedPartialSignatureMessage) error
	ProcessPostConsensus(signedMsg *types.SignedPartialSignatureMessage) error
}
```

GalRogozinski:
For topics the first validator can be used or we organize our topics via clusters

GalRogozinski (changes requested):
1. ClusterRunner - one runner with aggregated message
2. p2p MsgValidation
3. New topics

sips/cluster_consensus.md
```diff
@@ -0,0 +1,225 @@
...
+	// ProcessMessage processes a message routed to this Cluster
+	ProcessMessage(msg *types.SSVMessage)
+}
+
+type Cluster struct {
+	ClusterRunners     map[spec.Slot]ClusterRunner
+	Network           Network
+	Beacon            BeaconNode
+	OperatorID        OperatorID
+	ClusterShares     [spec.ValidatorPubKey]ClusterShares
```
y0sher, line 69:
```suggestion
	ClusterShares     map[spec.ValidatorPubKey]ClusterShares
```

sips/cluster_consensus.md
```diff
@@ -0,0 +1,225 @@
...
+	getSigner()           BeaconSigner
+}
+
+
+// ClusterRunner manages the duty cycle for a certain slot
+type ClusterRunner interface {
+	// Start the duty lifecycle for the given slot. Emits a message.
+    StartDuties(slot spec.Slot, attesterDuties []*types.Duty, syncCommit	teeDuties []*types.Duty) error
+	// Processes cosensus message
+    ProcessConsensus(consensusMessage *qbft.Message) ru error
```
y0sher, line 89: can you fix? there's some formatting issue

sips/cluster_consensus.md
```diff
@@ -0,0 +1,290 @@
...
+	// ProcessMessage processes a message routed to this Cluster
+	ProcessMessage(msg *types.SSVMessage)
+}
+
+type Cluster struct {
+	ClusterRunners     map[spec.Slot]ClusterRunner
+	Network           Network
+	Beacon            BeaconNode
+	OperatorID        OperatorID
+	ClusterShares     map[spec.ValidatorPubKey]ClusterShares
```
moshe-blox:
```suggestion
	Shares     map[spec.ValidatorPubKey]ClusterShares
```

sips/cluster_consensus.md
```diff
@@ -0,0 +1,296 @@
...
+	// Start the duty lifecycle for the given slot. Emits a message.
+    StartDuties(slot spec.Slot, attesterDuties []*types.Duty, syncCommitteeDuties []*types.Duty) error
+	// Processes cosensus message
+    ProcessConsensus(consensusMessage *qbft.Message) error
+	// Processes a post-consensus message
+	ProcessPostConsensus(msg ClusterSignaturesMessage) 
+}
+
+type ClusterRunner struct {
+	Share          *types.Share
```
y0sher, line 95: is single share enough? don't we need share per validator?

sips/cluster_consensus.md
```diff
@@ -0,0 +1,296 @@
...
+3.  Once `ProcessConsensus` decides, the `Cluster` will create a post consensus of PartialSignatureMessage that aggregates the beacon partial signature for all relevant validators.
+4. `Cluster` will process post-consensus partial signature messages and submit a beacon message for each validator.
+
+### Stopping Runs
+
+Previously we have letted new validator duties stop the run for the previous duty. For a cluster this is not a good idea and instead we will count on a tuned `CutOffRound` per duty to stop the instance.
+
+### Omitting Partial Signatures
+
+If in post-consensus stage for attestation duty, the duty's slot is lower than a validator's `ClusterShare's` `highestAttestingSlot`, the `ClusterRunner` will omit the post-consesnus message for this validator.
```
y0sher, line 150: IMO this description is lacking what happens regards a quorum, and/or who created the message. if we got 2 partials and we created the third should we still omit it? if we have 2 and received a third should we omit it?

sips/cluster_consensus.md
```diff
@@ -0,0 +1,296 @@
...
+
+#### Sync Committee
+`CutOffRound = 4  \\ one slot`
+
+#### Attestation
+`CutOffRound = 12 \\ one epoch`
+
+### ClusterID
+
+An identifier for cluster must be added to `MessageID`.
```
y0sher, line 160:
```suggestion
### MsgID + ClusterID
Today a MsgID is constructed from Validator and Duty type, in order to route messages into the cluster runners, identifier for cluster must be added. `ClusterID`.
```

sips/cluster_consensus.md
```diff
@@ -0,0 +1,296 @@
...
+
+```go
+[48]byte ClusterID
+
+// Return a 48 bytes ID for the cluster of operators
+func getClusterID(operatorIDs []OperatorID) ClusterID {
+	// Create a 16 bytes constant prefix for cluters
```
y0sher, line 167:
```suggestion
	// Create a 16 bytes constant prefix for clusters
```

sips/cluster_consensus.md
```diff
@@ -0,0 +1,296 @@
...
+sort.Slice(committee, func(i, j int) bool {
+	return committee[i] < committee[j]
+})
+// Convert to bytes
+bytes := make([]byte, len(committee)*4)
+for i, v := range committee {
+	binary.BigEndian.PutUint32(bytes[i*4:], uint32(v))
+}
+// Hash
+hashed := sha256.Sum256(bytes)
```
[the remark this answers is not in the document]
y0sher: isn't it the same? he is iterating over commitee

sips/cluster_consensus.md
```diff
@@ -0,0 +1,296 @@
...
+// Cluster is a cluster of a unique set of operators that run the same validator set
+type Cluster interface {
+	// Initializes and starts the runner for the duties for the given slot
+  	Start(slot spec.Slot, attesterDuties []*types.Duty, syncCommitteeDuties []*types.Duty) error
+	// ProcessMessage processes a message routed to this Cluster
+	ProcessMessage(msg *types.SSVMessage)
+}
+
+type Cluster struct {
+	ClusterRunners     map[spec.Slot]ClusterRunner
```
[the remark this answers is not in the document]
GalRogozinski: don't see the issue
in the key space you can have multiple slots

sips/cluster_consensus.md
```diff
@@ -0,0 +1,296 @@
...
+	getSigner()           BeaconSigner
+	// highestAttestingSlot holds the highest slot for which attester duty ran for a validator
+	highestAttestingSlot spec.Slot
+}
+
+
+// ClusterRunner manages the duty cycle for a certain slot
+type ClusterRunner interface {
+	// Start the duty lifecycle for the given slot. Emits a message.
+    StartDuties(slot spec.Slot, attesterDuties []*types.Duty, syncCommitteeDuties []*types.Duty) error
```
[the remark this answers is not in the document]
GalRogozinski: Cluster `Start` calls this

sips/cluster_consensus.md
GalRogozinski: This section only talks about omitting messages because of slots.

sips/cluster_consensus.md
```diff
@@ -0,0 +1,330 @@
...
+#### PartialSignatureMessages
+
+The current structure that we have in code can be unchanged.
+
+```go
+type PartialSignatureMessages struct {
+	Type     PartialSigMsgType
+	Slot     phase0.Slot
+	Messages []*PartialSignatureMessage
+}
```
MatheusFranco99, line 143: So, indeed, the share's signatures will be merged into the same message, right? If so, what will be the maximum number of signatures per message?

sips/cluster_consensus.md
```diff
@@ -0,0 +1,330 @@
...
+	PartialSignature Signature `ssz-size:"96"` // The Beacon chain partial Signature for a duty
+	SigningRoot      [32]byte  `ssz-size:"32"` // the root signed in PartialSignature
+    Signer           OperatorID // TODO: there is an optimization where we don't have to duplicate this field
+}
+```
+
+We must have the following flow:
+1. When you `calculateExpectedRootsAndDomain` (similar to `sync_committee_aggregator`). Use the `duty` objects to reconstruct the proper beacon data objects (i.e. `Attestation`).
+2. The above calculation can be used to create a mapping of `ValidatorIndex` to `root`
+3. When processing the messages find all the roots that have quorums, mark them, and sumbit corresponding beacon data to beacon chain.
```
MatheusFranco99, line 156: So, for attestation duties, we can relate the validators to the signing root. But what about the signatures for the sync committee duties? The signing root should be the same if I'm not mistaken.

GalRogozinski: It isn't the same because we sign Beacon Objects

GalRogozinski: https://github.com/bloxapp/ssv-spec/blob/ebb60143f17ea25edf09edeb398559d764b0e01b/ssv/sync_committee.go#L138-L143

GalRogozinski: you sign the head root for sync committee
for attestation it is the head root + source + target

sips/cluster_consensus.md
```diff
@@ -0,0 +1,356 @@
...
+  - We can no longer limit a single validator to two attestation attempts per epoch. If we were to count only attestation, we could limit a cluster with $V$ validators to do $2 \times V$ consensus per epoch. However, the fact the sync committee duties use the same consensus instances as attestations can force us to tolerate 32 consensus instances from a cluster in an epoch. Thus, we suggest limiting by $2 \times V$ executions only with a condition check that no validators of such cluster are doing sync committee duties in the epoch.
+  - For the `BNRoleAttesterOrSyncCommittee` role, the number of signatures in a `PartialSignatureMessages` is limited to $min(2 * V, V+SYNC_COMMITTEE_SIZE.)$.
+
+Regarding implementation,
+  - The `ConsensusState` is currently mapped by a `ConsensusID` that uses the validator public key and the role from the `MessageID`. Since `MessageID` will have a ClusterID, instead of a validator public key, for attestations and sync committees, the mapping of `ConsensusState` will need to change. We suggest either changing `ConsensusID` to encompass also a `ClusterID` or simply mapping `ConsensusState` by `MessageID`.
```
GalRogozinski, line 329: Message Validation section is fine by me..
But this specific section needs to be deleted before the SIP is merged

MatheusFranco99: Do you think this part regarding implementation should be removed from the SIP?
For me, no problem

sips/cluster_consensus.md
```diff
@@ -0,0 +1,362 @@
...
+type Cluster interface {
+	// Initializes and starts the runner for the duties for the given slot
+  	Start(duty Duty) error
+	// ProcessMessage processes a message routed to this Cluster
+	ProcessMessage(msg *signedSSVMessage *types.SignedSSVMessage)
+}
+
+type Cluster struct {
+	Runners     map[spec.Slot]ClusterRunner
+    Shares      map[ValidatorPubkey]Share
```
lior-blox, line 72: @GalRogozinski align please

sips/cluster_consensus.md
```diff
@@ -0,0 +1,362 @@
...
+    
+}
+
+type QBFTParams interface {
+    Operators()        []OperatorID
+    Quorum()           uint64
+    PartialQuorum() 	  uint64
+}
+
+=
```
lior-blox, line 86:
```suggestion

```

sips/cluster_consensus.md
```diff
@@ -0,0 +1,362 @@
...
+    StartDuty(duty Duty) error
+	// Processes cosensus message
+    ProcessConsensus(consensusMessage *qbft.Message) error
+	// Processes a post-consensus message
+    ProcessPostConsensus(msg PartialSignatureMessages) 
+}
+
+type ClusterRunner struct {
+	Shares          map[ValidatorPubkey]Share
+    QBFTParams     QBFTParams
```
lior-blox, line 105: align please

sips/cluster_consensus.md
```diff
@@ -0,0 +1,362 @@
...
+We must have the following flow:
+1. When you `calculateExpectedRootsAndDomain` (similar to `sync_committee_aggregator`). Use the `duty` objects to reconstruct the proper beacon data objects (i.e. `Attestation`).
+2. The above calculation can be used to create a mapping of `ValidatorIndex` to `root`
+3. When processing the messages find all the roots that have quorums for a certain validator, mark them, and sumbit corresponding beacon data to beacon chain.
+4. For the next message received attempt to complete quorum for other roots.
+
+
+#### Happy Flow
+
+1. `Cluster` receives duties that match a certain slot. If the slot is higher then `highestDecidedSlot` starts consensus for the relevant roles, and initializes a `ClusterRunner` for the relevant Validators.
```
lior-blox, line 175: @GalRogozinski I think we dont use highestDecidedSlot anymore, we should start consensus and then not broadcasting partial sig for a validator in case the slot is not higher than HighestAttestingSlot, correct?

sips/cluster_consensus.md
GalRogozinski: Max validators an operator can hold times 2 I believe

sips/cluster_consensus.md
GalRogozinski: only "Regarding Implementation"

sips/cluster_consensus.md
GalRogozinski: ok

sips/cluster_consensus.md
```diff
@@ -0,0 +1,440 @@
...
+
+Regarding syntax, the rules can stay the same.
+
+Regarding semantics,
+  - If `SSVMessage.MsgID` contains a CommitteeID (i.e. the role is attestation + sync committee) and if the CommitteeID doesn't exist in the current network, it should ignore the message.
+  - If a `ValidatorIndex` in `SignedPartialSignatureMessage.Message.Messages` is incorrect, considering the ValidatorPublicKey or the CommitteeID in the MessageID, it should ignore the message.
+
+Regarding duties' general rules,
+  - The duplicated partial signature rule (same validator, slot and type but different signing root or signature) now applies only to the Proposal, Aggregate, Sync Committee Contribution, Validator Registration and Voluntary Exit duties (since attestation and sync committee partial signature messages may include the same signing root.)
+  - If the same `ValidatorIndex` appears more than 2 times in `SignedPartialSignatureMessage.Message.Messages`, the message is rejected.
```
GalRogozinski, line 405: @MatheusFranco99 
> since attestation and sync committee partial signature messages may include the same signing root.

you sure about this?

> If the same `ValidatorIndex` appears more than 2 times in `SignedPartialSignatureMessage.Message.Messages`, the message is rejected.

if we have attestation and sync committee will this hold?
I think maybe this part is due to an older version of the SIP

sips/cluster_consensus.md
```diff
@@ -0,0 +1,440 @@
...
+
+Regarding duties' specific rules,
+  - The attestation's and sync committee's specific rules are dropped in favor of a common set of rules for the new duty type attestation + sync committee. Namely, the higher attestation limits are used (34 slots, 12 rounds).
+  - We can no longer limit a single validator to two attestation attempts per epoch. If we were to count only attestation, we could limit a committee with $V$ validators to do $2 \times V$ consensus per epoch. However, the fact the sync committee duties use the same consensus instances as attestations can force us to tolerate 32 consensus instances from a committee in an epoch. Thus, we suggest limiting by $2 \times V$ executions only with a condition check that no validators of such committee are doing sync committee duties in the epoch.
+  - For the `BNRoleAttesterOrSyncCommittee` role, the number of signatures in a `PartialSignatureMessages` is limited to $min(2 * V, V+SYNC\_COMMITTEE\_SIZE.)$.
```
GalRogozinski, line 410:
```suggestion
  - For the `RunnerRole` role, the number of signatures in a `PartialSignatureMessages` is limited to $min(2 * V, V+SYNC\_COMMITTEE\_SIZE.)$.
```

sips/cluster_consensus.md
MatheusFranco99: Yes, if I'm not mistaken, two validators may be signing the same AttestationData (
```go
type AttestationData struct {
	Slot            Slot
	Index           CommitteeIndex
	BeaconBlockRoot Root
	Source          *Checkpoint
	Target          *Checkpoint
}
```
) if they belong to the same committee and agree on the vote (with the same `DomainAttester` domain.

Regarding the "2 times", the idea is that at most a validator index is both doing the attestation and sync committee duties. So, it can have at most 2 signatures related to it.

sips/cluster_consensus.md
```diff
@@ -0,0 +1,440 @@
...
+ - A decay $d_2$
+ - A cap value ($cap_2$) for the counter
+ - A weight ($w_2$) to multiply the counter value and sum the result to the topic's score.
+
+The decay $d_2$ can stay the same (0.3162277660168379, which decays 1 to 0.01 in 4 epochs). The cap value is computed using the mesh target size, $D = 8$, and the expected message rate per decay interval, and it's defined as the convergence value if one keeps sending twice the number of messages it's expected to send. In other words,
+
+$$cap_2 = \frac{2m}{D} \times \frac{1}{(1-d_2)}$$
+
+The message rate for a validator per decay interval (one epoch), used to be calculated as $600/10000 \times (32 \times 12)$. To account for all validators in a topic, we multiply this value by $\frac{V}{32}$. Then, to account for the improvement of this change, considering that the message rate will drop to 15% of the current value, we multiply it by $0.15$ and, finally, we get
```
GalRogozinski, line 426: In your simulation didn't you say it will drop to 21.61%?

sips/cluster_consensus.md
```diff
@@ -0,0 +1,440 @@
...
+
+$$cap_2 = \frac{2}{D} \times \left( \frac{600}{10000} \times (32 \times 12) \times \frac{V}{128} \times 0.15 \right) \times \frac{1}{1-d_2}$$
```
GalRogozinski, line 428: In the description you still write $\frac{V}{32}$...
So actually before we had a mistake for not multiplying by this value?

sips/cluster_consensus.md
MatheusFranco99: New Mainnet estimation.

sips/cluster_consensus.md
MatheusFranco99: The description should also have V/128. Updated it

sips/cluster_consensus.md
```diff
@@ -0,0 +1,440 @@
...
+// CommitteeDuty implements Duty
+type CommitteeDuty struct {
+	Slot         spec.Slot
+	BeaconDuties []*BeaconDuty
+}
+
+// Committee is a committee of a unique set of operators that have shared validators
+type Committee interface {
+	// Initializes and starts the runner for the duties for the given slot
+  	StartDuty(duty Duty) error
```
lior-blox, line 65: @GalRogozinski should be CommitteeDuty, correct?

sips/cluster_consensus.md
```diff
@@ -0,0 +1,440 @@
...
+	Operator                types.Operator
+	SignatureVerifier       types.SignatureVerifier
+	CreateRunnerFn          func() *CommitteeRunner
+	// Save the highest slot a validator attested to
+	HighestAttestingSlotMap map[types.ValidatorPK]spec.Slot
+}
+
+// StartDuty starts a new duty for the given slot
+func (c Committee) StartDuty(duty types.CommitteeDuty) error {
+	// do we need slot?
```
lior-blox, line 81: @GalRogozinski resolve comment

sips/cluster_consensus.md
```diff
@@ -0,0 +1,440 @@
...
+  - If the same `ValidatorIndex` appears more than 2 times in `SignedPartialSignatureMessage.Message.Messages`, the message is rejected.
+
+Regarding duties' specific rules,
+  - The attestation's and sync committee's specific rules are dropped in favor of a common set of rules for the new duty type attestation + sync committee. Namely, the higher attestation limits are used (34 slots, 12 rounds).
+  - We can no longer limit a single validator to two attestation attempts per epoch. If we were to count only attestation, we could limit a committee with $V$ validators to do $2 \times V$ consensus per epoch. However, the fact the sync committee duties use the same consensus instances as attestations can force us to tolerate 32 consensus instances from a committee in an epoch. Thus, we suggest limiting by $2 \times V$ executions only with a condition check that no validators of such committee are doing sync committee duties in the epoch.
```
lior-blox, line 409: @MatheusFranco99 Why we allow 2 *v consensuses per epoch?
Does that mean that If a committee has 500 validators we allow up to 1000 consensuses?

sips/cluster_consensus.md
GalRogozinski: `Duty` is an interface
But perhaps in the SIP it is indeed best to use `CommitteeDuty` for clarity

I will change

sips/cluster_consensus.md
MatheusFranco99: @lior-blox the limit would be min(32, 2*v). So for any $v>=16$, the maximum limit will always be 32 consensuses per epoch

sips/committee_consensus.md
```diff
@@ -0,0 +1,486 @@
...
+			if slot < beaconDuty.Slot {
+				validatorsToStop[beaconDuty.Slot] = validatorPK
+				slot = beaconDuty.Slot
+			} else { // else don't run duty with old slot
+				duty.BeaconDuties[i] = nil
+			}
+		}
+	}
+	return duty, validatorsToStop, slotMap
+}
```
moshe-blox, line 246: won't this stop attester duties after one slot for validators who currently have a sync duty?

sips/committee_consensus.md
```diff
@@ -0,0 +1,486 @@
...
+	Type     PartialSigMsgType
+	Slot     phase0.Slot
+	Messages []*PartialSignatureMessage
+}
+
+// PartialSignatureMessage is a msg for partial Beacon chain related signatures (like partial attestation, block, randao sigs)
+type PartialSignatureMessage struct {
+	PartialSignature Signature `ssz-size:"96"` // The Beacon chain partial Signature for a duty
+	SigningRoot      [32]byte  `ssz-size:"32"` // the root signed in PartialSignature
+	Signer         OperatorID
```
moshe-blox, line 195: i think this can be removed since it's already in the SignedSSVMessage, right?

if this would be a big change, then maybe we could just move this field up to the PartialSignatureMessages struct, just so it won't repeat for every signature

sips/committee_consensus.md
```diff
@@ -0,0 +1,486 @@
...
+
+In order to route consensus messages to the correct consensus runner, the `CommitteeID` field will be included in the `MessageID` replacing `ValidatorPublicKey`.
+
+### Role
+
+This is used to route the message to the correct runner. Since `CommitteeRunner` has several beacone roles we will create a new `RunnerRole` in the MessageID.
+
+```go
+// TODO since this is on wire no real need to take 32 bits
+type RunnerRole int32
```
moshe-blox, line 301: yes, this can be changed to int8, but if it's not a very trivial change, i'd not include it in this SIP, because i doubt this optimization would be even visible in any chart at all (be it traffic, memory usage, etc..)

sips/committee_consensus.md
```diff
@@ -0,0 +1,486 @@
...
+
+func (msg MessageID) GetSenderID() []byte {
+	return msg[senderIDStartPos : senderIDStartPos+senderIDSize]
+}
+
+func parseMessageID(messageID MessageID) (DomainType, BeaconRole, []byte) {
+	domain := messageID.GetDomain()
+	role := messageID.GetRoleType()
+	if role == BNRoleSyncCommitteeContribution {
+		return domain, BNRoleAttesterOrSyncCommittee, messageID.GetSenderID()[16:]
```
moshe-blox, line 351: if there's no special reason for putting committee ID at the end of sender ID, i'd recommend the other way around due to it being less of a "surprise" to the common pattern of taking up the left side of bigger than needed arrays.

plus, it'd make such conversions correct by default (no magic number is required):

```go
cid := CommitteeID(senderID)
```

sips/committee_consensus.md
```diff
@@ -0,0 +1,486 @@
...
+			Index:           math.MaxUInt64,
+			BeaconBlockRoot: bv.BlockRoot,
+			Source:          bv.Source,
+			Target:          bv.Target,
+		}
+
+		for _, sharePublicKey := range sharePublicKeys {
+			if err := signer.IsAttestationSlashable(sharePublicKey, attestationData); err != nil {
+				return errors.Wrap(err, "slashable attestation")
+			}
```
moshe-blox, line 405: unsure if this is final, but won't this mean 1 slashable validator fails the entire committee?

moshe-blox (changes requested):

sips/committee_consensus.md
MatheusFranco99: Yes, I've already fixed this issue in implementation.
@GalRogozinski

sips/committee_consensus.md
MatheusFranco99: I think we all want to remove it. I don't remember the reason now why it's still there
Was there any? @GalRogozinski

GalRogozinski: Just so that old runners don't change to reduce the amount of changes and risk for bugs in a rushed release...
you guys want to change it now?

sips/committee_consensus.md
GalRogozinski: yes, but:
1. maybe it is ok we find a new leader?
2. I was trying to make a concrete example of how we can be at a deadlock and failed... maybe we can, but it requires more research

sips/committee_consensus.md
GalRogozinski: removing the TODO because there is a comment..
If we don't do this it should be added to #39

GalRogozinski: for you it is trivial @moshe-blox ?

sips/cluster_consensus.md
GalRogozinski: This can happen more likely because of a delay of duty processing rather than a fork

sips/committee_consensus.md
GalRogozinski: This won't work due to 0x00 prefix

sips/committee_consensus.md
```diff
@@ -0,0 +1,480 @@
...
+	Type     PartialSigMsgType
+	Slot     phase0.Slot
+	Messages []*PartialSignatureMessage
+}
+
+// PartialSignatureMessage is a msg for partial Beacon chain related signatures (like partial attestation, block, randao sigs)
+type PartialSignatureMessage struct {
+	PartialSignature Signature `ssz-size:"96"` // The Beacon chain partial Signature for a duty
+	SigningRoot      [32]byte  `ssz-size:"32"` // the root signed in PartialSignature
+	Signer         OperatorID
```
alonmuroch, line 188: this can be removed in a later stage, @GalRogozinski please add an issue for it

sips/committee_consensus.md
```diff
@@ -0,0 +1,480 @@
...
+We must have the following flow:
+1. Calculate `expectedRootsAndBeaconObjects()`. Use the `duty` objects to reconstruct the proper beacon data objects (i.e. `Attestation`).
+2. The above calculation can be used to create a mapping of `ValidatorIndex` to `root`, and `root` to `beaconObject`.
+3. When processing the messages find all the roots that have quorums for a certain validator, mark them, and submit corresponding beacon data to beacon chain.
+4. For the next message received attempt to complete quorum for other roots.
+
+
+#### Happy Flow
+
+1. `Committee` receives duties that match a certain slot. Then `StartDuty` and filter the validators who have a higher `HighestAttestingSlot`. Initialize a `CommitteeRunner` for the relevant Validators and slot. For each validator that we start, mark its old beacon duties as stopped.
```
alonmuroch, line 202: I don't see it as a security concern as slashing protection protects against it. Should be on implementation side

sips/committee_consensus.md
```diff
@@ -0,0 +1,480 @@
...
+### Stopping Runs
+
+Previously we have letted new validator duties stop the run for the previous duty. Now we don't stop the runner but just mark certain validators as stopped and omit the partial sig emission for them.
+
+### Omitting Partial Signatures
+
+If in post-consensus stage for attestation duty, the duty's slot is lower than a validator's `highestAttestingSlot`, the `CommitteeRunner` will omit the post-consesnus message for this validator. 
+
+### Cutoff Round
+We will also create a better Cutoff Round.
```
alonmuroch, line 257: need explanation why we change it

sips/committee_consensus.md
```diff
@@ -0,0 +1,480 @@
...
+	RoleValidatorRegistration
+	RoleVoluntaryExit
+
+	RoleUnknown = -1
+)
+```
+
+### MessageID
+
+`ValidatorPublicKey` and `CommitteeID` will be used interchangeably in `MessageID`. `Role` will change location because it is used to determine ID type.  `CommitteeID` is 32 bytes long so it will be encoded with a `0x00` 16 bytes prefix to make it the same length as `ValidatorPublicKey`.
```
alonmuroch, line 312: ValidatorPublicKey remains here because we leave the validator struct in tact, later when we remove it we will need another fork
@GalRogozinski please open an issue

sips/committee_consensus.md
```diff
@@ -0,0 +1,480 @@
...
+// Committee runner implements the old runner interface
+type CommitteeRunner struct {
+	// Important fields only
+	Shares          map[ValidatorPubkey]Share
+    Operator        Operator
+	QBFTController *qbft.Controller
+	BeaconNetwork  *types.BeaconNetwork
+}
+
+// BeaconVote is the consensus data
```
alonmuroch, line 143: better explanation

sips/committee_consensus.md
```diff
@@ -0,0 +1,480 @@
...
+	return domain, role, messageID.GetSenderID()
+}
+```
+
+
+### Consensus Data
+
+We note that the data needed for a consensus execution is the same for all validators. Thus, we can create a `BeaconVote` object that will hold the data for all validators. The data needed for the `SyncCommittee` role is a subset of the data needed for the `Attestation` role. Thus we can always query the beacon node for attestation data and pass this data to relevant runners.
+
+`ConsensusData` currently holds the `duty` field to make sure that all committee members agree on committee information. However, if there is a difference between committee members there is no guarantee that a run will be triggered on the first place. This is because different views may cause [different shuffles](https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#compute_committee). Therefore we can rely on local view of `duty`, and keep the field empty.
```
alonmuroch, line 356: need to re-write this, explain consensus data is the old struct and is left because of the old runners

sips/committee_consensus.md
```diff
@@ -0,0 +1,480 @@
...
+
+		if bv.Target.Epoch > estimatedCurrentEpoch+1 {
+			return errors.New("attestation data target epoch is into far future")
+		}
+
+		if bv.Source.Epoch >= bv.Target.Epoch {
+			return errors.New("attestation data source >= target")
+		}
+
+		attestationData := &phase0.AttestationData{
```
alonmuroch, line 385: let's separate it to a different function because we might have multiple checks (not just attestation)

alonmuroch (changes requested):

sips/committee_consensus.md
```diff
@@ -0,0 +1,480 @@
...
+The `Committee` will hold a `CommitteeRunner` object for each slot it is running a beacon duty for.
+
+For other duty roles the old design will remain.
+
+#### Code
+
+```go
+// Share holds all info about the validator share
+// All the operator related data moved to operator
+type Share struct {
```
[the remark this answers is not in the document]
GalRogozinski: The current hierarchy had to be this way to work well with `BaseRunner`... changing names

sips/committee_consensus.md
```diff
@@ -0,0 +1,480 @@
...
+
+// Committee is a committee of a unique set of operators that have shared validators
+type Committee interface {
+	// Initializes and starts the runner for the duties for the given slot
+  	StartDuty(duty CommitteeDuty) error
+	// ProcessMessage processes a message routed to this Committee
+	ProcessMessage(msg *types.SignedSSVMessage)
+}
+
+type Committee struct {
```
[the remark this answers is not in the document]
GalRogozinski: in v2

sips/committee_consensus.md
```diff
@@ -0,0 +1,480 @@
...
+	Runners                 map[spec.Slot]*CommitteeRunner
+	Operator                types.Operator
+	SignatureVerifier       types.SignatureVerifier
+	CreateRunnerFn          func() *CommitteeRunner
+	// Save the highest slot a validator attested to
+	HighestAttestingSlotMap map[types.ValidatorPK]spec.Slot
+}
+
+// Duty interface
+type Duty interface {
```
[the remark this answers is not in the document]
GalRogozinski: no interfaces in SIP

sips/committee_consensus.md
```diff
@@ -0,0 +1,480 @@
...
+}
+
+// CommitteeDuty aggregates attesting and sync committee duties
+type CommitteeDuty struct {
+	Slot         spec.Slot
+	BeaconDuties []*BeaconDuty
+}
+
+// CommitteeRunner manages the duty cycle for a certain slot
+type CommitteeRunner interface {
```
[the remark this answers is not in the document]
GalRogozinski: ok

sips/committee_consensus.md
```diff
@@ -0,0 +1,480 @@
...
+Again with Monte Carlo simulations using the Mainnet dataset, the number of attestation duties aggregated presented the following distribution.
+
+<p align="center">
+<img src="./images/cluster_consensus/aggregated_duties.png"  width="50%" height="10%">
+</p>
+
+## Spec changes
+
+### Design
+
```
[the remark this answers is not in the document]
GalRogozinski: shouldn't be part of this SIP imo

sips/committee_consensus.md
GalRogozinski: part of https://github.com/ssvlabs/ssv-spec/issues/368

sips/committee_consensus.md
GalRogozinski: ok

sips/committee_consensus.md
GalRogozinski: only attestation is slashable
I don't see anything to fix

sips/committee_consensus.md
GalRogozinski: we need to make sure that in spec we have slashing check enabled before we sign partial sigs even if they pass consensus

sips/committee_consensus.md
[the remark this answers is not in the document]
GalRogozinski: will be done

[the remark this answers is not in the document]
alonmuroch: It should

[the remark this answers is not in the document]
GalRogozinski: done

sips/committee_consensus.md
GalRogozinski: there is a actually an issue...
you need both signer and validator index now since different validators in the committee may sign the same root

sips/committee_consensus.md
```diff
@@ -0,0 +1,485 @@
...
+Therefore, to compute the message rate for a certain topic, we need to get, as input, the list of committees and validators present in such a topic and adjust the calculation as the following script:
+
+```go
+// Expected number of committee duties per epoch due to attestation beacon duties given
+// the number of validators in the committee
+func expectedNumberOfCommitteeDutiesPerEpochDueToAttestation(numValidators int) float64 {
+	k := float64(numValidators)
+	n := 32.0
+	return n * (1 - math.Pow((n-1)/n, k))
+}
```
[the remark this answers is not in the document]
GalRogozinski: It is actually a bit unclear to me why your way is correct and not something that is more similar to the end of this article about the [coupons collector problem](https://towardsdatascience.com/coupon-collectors-problem-a-probability-masterpiece-1d5aed4af439)

Let use notation E(k) where k is the number of validators in committee
The expected value should be:
E(1) = 1
E(2) = E(1) + (n-1)/n
E(3) = E(2) + (n-2)/n

no?

GalRogozinski (changes requested):
just clarify to me your calculation please...
maybe I am wrong, but I am not sure why

sips/committee_consensus.md
```diff
@@ -0,0 +1,485 @@
...
+ - A decay $d_2$
+ - A cap value ($cap_2$) for the counter
+ - A weight ($w_2$) to multiply the counter value and sum the result to the topic's score.
+
+The decay $d_2$ can stay the same (0.3162277660168379, which decays 1 to 0.01 in 4 epochs). The cap value is defined as the convergence value if one keeps sending twice the number of messages it's expected to send. Thus, it's computed using the mesh target size, $D = 8$, and the expected message rate per decay interval. In other words,
+
+$$cap_2 = \frac{2m}{D} \times \frac{1}{(1-d_2)}$$
+
+The message rate for a validator per decay interval (one epoch), used to be calculated as $600/10000 \times (32 \times 12)$. To account for all validators in a topic, we used to multiply this value by $\frac{V}{128}$. However, now, the number of validators in a topic is not uniform (so we can't use the $\frac{V}{128}$ estimation) and the message rate should be regarded as "by committee" instead of "by validator" since the number of messages depends on the aggregation of the committee's validators duties.
```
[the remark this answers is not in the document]
MatheusFranco99: 600/10000 is a "messages per second per validator" metric from the stage. However, notice that it's not used any longer.

sips/committee_consensus.md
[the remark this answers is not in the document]
GalRogozinski: coupons collector problem is a bit different than this problem... I got confused....
I looked at it again and your formula is correct

GalRogozinski (approved):

sips/committee_consensus.md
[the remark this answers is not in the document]
MatheusFranco99: It's similar indeed. The difference is thin
Nice

sips/committee_consensus.md
```diff
@@ -0,0 +1,505 @@
...
+
+```go
+	AlanMainnet = DomainType{0x0, 0x0, MainnetNetworkID.Byte(), 0x1}
+```
+
+#### Code
+
+```go
+// Share holds all info about the validator share
+// All the operator related data moved to operator
```
y0sher, line 63:
```suggestion
// NOTE: All the operator related data moved to operator
```

sips/committee_consensus.md
```diff
@@ -0,0 +1,505 @@
...
+}
+
+// ShareMember holds ShareValidatorPK and ValidatorIndex
+type ShareMember struct {
+	SharePubKey ShareValidatorPK
+	Signer      OperatorID
+}
+
+// Operator represents an SSV operator node that is part of a committee
+type CommitteeMember struct {
```
y0sher, line 81:
```suggestion
// CommitteeMember represents an SSV operator node that is part of a committee
type CommitteeMember struct {
```

sips/committee_consensus.md
moshe-blox: at this point it's not worth the change imo

sips/committee_consensus.md
moshe-blox: we agreed that at this point this should remain unchanged and open to discussions in the future

sips/committee_consensus.md
y0sher: IMO its best to change in this now than doing another fork for this

sips/committee_consensus.md
```diff
@@ -0,0 +1,480 @@
...
+}
+
+// CommitteeDuty implements Duty
+type CommitteeDuty struct {
+	Slot         spec.Slot
+	BeaconDuties []*BeaconDuty
+}
+
+// Committee is a committee of a unique set of operators that have shared validators
+type Committee interface {
```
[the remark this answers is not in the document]
GalRogozinski: interfaces removed

sips/committee_consensus.md
```diff
@@ -0,0 +1,505 @@
...
+
+### Cutoff Round
+Previously new duties stopped consensus intsances. We cannot do that anymore since duties don't neccesarily map to specific validators. So we create a better Cutoff Round that will stop the consensus instance in a more sensible time. If the Sync Committee beacon message is longer than a slot tehn it won't enter the beacon chain.
+An Attestation has at most 2 epochs to be included. However since after round 8 we use a slow timeout, if after 4 additional change rounds the instance didn't decide then the chances of it deciding are very low.
+
+#### Sync Committee
+`CutOffRound = 4  \\ one slot`
+
+#### Attestation
+`CutOffRound = 12 \\ one epoch`
```
moshe-blox, line 217: Do we actually pay attention to the different cutoff round now that both duty types run the same consensus?

GalRogozinski: We should use 12 as a global cutoff attestation round for all

y0sher (approved):

sips/committee_consensus.md
GalRogozinski: Opened issue here: https://github.com/ssvlabs/ssv-spec/issues/422
we may do it as part of the next fork

sips/committee_consensus.md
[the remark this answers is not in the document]
GalRogozinski: https://github.com/ssvlabs/ssv-spec/issues/451

PR #45, opened 2024-03-15
Reduce signatures - SIP

MatheusFranco99:
This SIP attempts to reduce the current cost of signature verification. Specifically, it removes the usage of BLS where it's not needed, such as in consensus messages and the upper type of partial signature messages. To accomplish that, it drops the current types for consensus and partial signature messages that use BLS in favor of a common wrapper with RSA signatures.

### DIFF
--- a/sips/reduce_signatures_fork.md
+++ b/sips/reduce_signatures_fork.md
@@ -1,0 +1,152 @@
+|     Author     |          Title           | Category |       Status        |    Date    |
+| -------------- | ------------------------ | -------- | ------------------- | ---------- |
+| Matheus Franco | Reduce Signatures - Fork | Core     | open-for-discussion | 2024-03-15 |
+
+## Summary
+
+Move the `FullData` of the `SignedMessage` type into the `Message` type and use a common wrapper for consensus and partial signature messages.
+
+## Motivation
+
+Reduce the cryptography cost by replacing BLS with RSA as the common wrapper signature type.
+
+## Rationale
+
+Currently, the hierarchy of message structures is as follows.
+
+```mermaid
+flowchart TB
+	SignedSSVMessage
+	SSVMessage
+	SignedMessage
+	Message
+	SignedPartialSignatureMessage
+	PartialSignatureMessages
+	PartialSignatureMessage
+
+	SignedSSVMessage --> SSVMessage
+	SSVMessage --> SignedPartialSignatureMessage
+	SignedPartialSignatureMessage --> PartialSignatureMessages
+	PartialSignatureMessages --> PartialSignatureMessage
+	SSVMessage --> SignedMessage
+	SignedMessage --> Message
+```
+
+The `SignedSSVMessage`, `SSVMessage`, `SignedMessage` and `SignedPartialSignatureMessage` types are shown below.
+
+```go
+type SignedSSVMessage struct {
+	OperatorID OperatorID
+ 	Signature  []byte // RSA Signature
+ 	Data       []byte
+}
+type SSVMessage struct {
+	MsgType MsgType
+	MsgID   MessageID
+	Data 	[]byte
+}
+type SignedMessage struct {
+	Signature 	types.Signature // BLS Signature
+	Signers   	[]types.OperatorID
+	Message 	Message
+	FullData 	[]byte
+}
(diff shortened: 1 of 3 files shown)


GalRogozinski (changes requested):
Looks nice

2 things:
1. Didn't understand how\why you get rid of BLS for PartialSigMessages? In the discussion I just talked about QBFT messages
2. It seems to me like Decided messages should be an inseparable part of this SIP.
We need to say a word about them

btw, put a link to the discussion please

MatheusFranco99:
@GalRogozinski 
1. I probably understood it wrong from the discussion. I thought that we would change the `FullData` place to use a common wrapper for Consensus and PartialSig messages with RSA instead of BLS. What did it actually mean?
2. Added a section for *decided* messages. I also talk about it in the last part of "Message Structures" and in the last part of "Message Validation". 
Thx for the feedback!

GalRogozinski:
I just meant for QBFT

GalRogozinski:
@MatheusFranco99 
actually maybe what you did is good I think I got confused :-P

GalRogozinski (approved):
Looks good!

The last section again reminded me why it is important that we eventually add MsgValidation to spec

moshe-blox (approved):

sips/reduce_signatures_fork.md
```diff
@@ -0,0 +1,163 @@
...
+}
+```
+
+To handle *decided* messages, we also propose allowing multiple signers and signatures in the `SignedSSVMessage`.
+
+
+```go
+type SignedSSVMessage struct {
+	OperatorID []OperatorID
+ 	Signature  [][]byte
```
y0sher, line 119: why is this a slice of slices?

sips/reduce_signatures_fork.md
```diff
@@ -0,0 +1,163 @@
...
+
+Except by messages inside justifications, the consensus module should not verify the signature of the received `SignedMessage` since it assumes that the message validation module already performed this check.
+
+Every sub-processing function of the consensus module, such as the `UponProposal`, `UponPrepare`, etc, should also receive a `SignedSSVMessage` instead of the deprecated `SignedMessage`.
+
+The message container must also stop storing `SignedMessage` objects in favor of storing `SignedSSVMessage` objects.
+
+### Partial signature message processing
+
+Currently, the partial signature processing functions accept a `SignedPartialSignatureMessage`. Now, it should accept a `SignedSSVMessage`. Similarly to the consensus module, it should assume that the message validation module already verified the RSA signature.
```
y0sher, line 145: latest changes are validating RSA signature inside the validator struct and not msg validation.

sips/reduce_signatures_fork.md
```diff
@@ -0,0 +1,163 @@
...
+
+### Runner
+
+Aligned to the changes on the `Validator` module, the `Runner` module should accept only `SignedSSVMessage` messages to its processing functions.
+
+### Message Validation
+
+Message validation should stop decoding the `SignedSSVMessage` to the deprecated `SignedMessage` or `SignedPartialSignatureMessage`. To check the content of the message it must decode the message data directly to one of the `Message` or `PartialSignatureMessages` types.
+
+Message validation should only allow multiple signers in the `SignedSSVMessage` message for *decided* messages (a check that was previously done for the `SignedMessage` type). It should verify it according to the [description above](#decided-messages).
```
y0sher, line 159: knowledge base and rules should be updated after this is merged.

GalRogozinski (commented):
A little observation

When we put `SignedSSVMessage` in a justification we save more data than we need...
Not sure how to resolve this and currently we shouldn't change anything...

just something to think of for the next fork

sips/reduce_signatures_fork.md
MatheusFranco99: Yes, it will. Thanks!

MatheusFranco99:
@GalRogozinski What do you mean by "we save more data than we need"?

GalRogozinski:
> What do you mean by "we save more data than we need"?

We save `SSVMessage` that includes routing data not needed for Consensus.
Since we verify the signatures before we hit (consensus) Instance processing and also after it is hard to model the message structure well

PR #46, opened 2024-03-19
Partial Signature Verification Aggregation

MatheusFranco99:
This SIP attempts to reduce the current cost of signature verification. Specifically, it boosts the *partial signature* steps,  in which a quorum is needed to trigger some action. To accomplish that, it proposes performing a single BLS verification on the reconstructed signatures only when a quorum is reached.

The SIP is based on discussion https://github.com/bloxapp/SIPs/discussions/38.


sips/reduce_signatures_no_fork.md
```diff
@@ -0,0 +1,129 @@
...
+Cryptography is, commonly, the biggest processing overhead in distributed systems. In our case, the BLS scheme is used which is also one of the heaviest cryptography protocols. Thus, performing one BLS verification for each in-committee message represents an important scalability barrier.
+
+## Rationale
+
+### 1st change
+
+A partial signature message implies one RSA verification (in the message validation layer) and two BLS verifications (for `SignedPartialSignatureMessage` and `PartialSignatureMessage`). Notice that two signatures are used to verify that the sender is correct and one is used to verify if the (same) sender correctly signed the beacon partial signature. Therefore, we can, simply, ignore the first BLS signature since the RSA signature is already being checked.
+
+> [!NOTE]
+> Since we want a no-fork change, one must still BLS-sign twice to be compatible with the previous version.
```
GalRogozinski, line 20: There is still a rogue key attack vector here ;-\

sips/reduce_signatures_no_fork.md
```diff
@@ -0,0 +1,129 @@
...
+The modification required in the consensus' *prepare* and *commit* functions is equivalent to the one shown in the [partial signature messages processing section](#partial-signature-messages-processing).
+
+## Drawbacks
+
+In the current version, we verify the BLS signature of each message once it is received and, thus, we also drop it immediately if it's wrong. So, if we were to receive two signatures from the same signer, we would store only the correct one (if any). With this proposal, we would need to decide what to do with the signatures, since we don't know which is correct. For that, we could follow one of the following approaches:
+1. verify the duplicated signatures and store the single correct one (if any).
+2. randomly drop one of them.
+3. store and try both until a quorum of signers is reached.
+
+The first approach implies signature verification (what we are trying to reduce), but it's the safest one and, thus, it should be the one adopted. Also, this represents an uncommon case and, therefore, the practical effects are negligible.
```
GalRogozinski, line 129: you can write a word that msg validation should prevent this, so it is best to raise an error

GalRogozinski (changes requested):

sips/reduce_signatures_no_fork.md
MatheusFranco99: How is that? Can you give an example?

sips/reduce_signatures_no_fork.md
MatheusFranco99: Msg validation already prevents a node from sending two diff partial signatures for the same duty step from the same signer. Though different nodes could sent two diff partial signatures for the same duty step from the same signer. But I think there's no way of stopping that without verifying the signatures.

sips/reduce_signatures_no_fork.md
GalRogozinski: Can't an attacker put a bad sig on the first BLS check simply?

MatheusFranco99: Yes, it can. But let's say one malicious operator creates an invalid signature on the `SignedPartialSignatureMessage`. What is the attack outcome in generating an inconsistent state?

lior-blox (approved):

GalRogozinski (approved):

moshe-blox (approved):

alonmuroch:
Any projected gains from this? in terms of speed/ CPU/ etc?

MatheusFranco99:
@alonmuroch 
This small change allows the cryptography cost to be reduced to 77% of the current value, a 1.3x boost.

PR #49, opened 2024-04-07
Cluster based consensus changes

GalRogozinski:
Add support for:
1. Single Cluster Runner
2. New ConsensusData
3. When to Stop Runs


sips/cluster_consensus.md
```diff
@@ -114,12 +166,20 @@ type PartialSigRunners map[ValidatorPublicKey]PartialSigRunner
...
-2. `Cluster` receives consensus messages and hands them over to the `QBFTController` that has unchanged logic.
-3. Once `ProcessConsensus` returns `decided = true`, the `Cluster` will call `UponDecided(cd)` for each `PartialSigRunner` that has been initialized. This will cause an omission of a partialSigMessage for each validator.
+1. `Cluster` receives duties that match a certain slot. If the slot is higher then `highestDecidedSlot` starts consensus for the relevant roles, and initializes a `ClusterRunner` for the relevant Validators.
+2. `Cluster` receives consensus messages and hands them over `ClusterRunner` that hands them over to the `QBFTController` that has unchanged logic. The only difference is `BeaconVote` is used as the ConsensusData object.
+3. Once `ProcessConsensus` decides, the `Cluster` will create a post consensus message like before.
 4. `PartialSigRunner` will process post-consensus partial signature messages as before.
 
+### Stopping Runs
+
+Previously we have letted new validator duties stop the run for the previous duty. For a cluster this is not a good idea and instead and instead we will count on a tuned `CutOffRound` per duty to stop the instance.
```
MatheusFranco99, line 176:
```suggestion
Previously we have letted new validator duties stop the run for the previous duty. For a cluster this is not a good idea and instead we will count on a tuned `CutOffRound` per duty to stop the instance.
```

MatheusFranco99:
Will the post-consensus signatures for all validators belong to a single message? Or there will be a message for each?

MatheusFranco99:
If the cluster receives a new `Start(duties)` call for slot X but hasn't finished the duties for slot X-1, what will happen?

GalRogozinski:
@MatheusFranco99 

1. will make it clear that we unite a single message
2. It will continue with the duty until `CutoffRound`. 
I think we need to make sure valueCheckF is called again when we sign the beacon object to ensure slashing protection

PR #51, opened 2024-05-21
SIP: topic by committee ID

MatheusFranco99:
This SIP aims to reduce the message rate processed by operators through the change of how validators are assigned to topics. Until now, a validator's assigned topic used to be computed using the `ValidatorPublicKey`. Instead, we propose using the `CommitteeID`.


GalRogozinski (approved):

sips/topic_by_committee_id.md
```diff
@@ -0,0 +1,97 @@
...
+	})
+	// Convert to bytes
+	bytes := make([]byte, len(committee)*4)
+	for i, v := range committee {
+		binary.LittleEndian.PutUint32(bytes[i*4:], uint32(v))
+	}
+	// Hash
+	return CommitteeID(sha256.Sum256(bytes))
+}
+topicID := binary.LittleEndian.Uint64(getCommitteeID(committee)) % subnetsCount
```
moshe-blox, line 51: Uint64 is only 8 bytes, so when converting from [32]byte to it, we lose the latter 24 bytes.

We use big.Int in the implementation to avoid this issue:
```go
func CommitteeSubnet(cid spectypes.CommitteeID) int {
	subnet := new(big.Int).Mod(new(big.Int).SetBytes(cid[:]), new(big.Int).SetUint64(subnetsCount))
	return int(subnet.Int64())
}
```

moshe-blox (changes requested):

sips/topic_by_committee_id.md
MatheusFranco99: Done

moshe-blox (approved):

sips/topic_by_committee_id.md
```diff
@@ -0,0 +1,101 @@
...
+// Return a 32 bytes ID for the committee of operators
+func getCommitteeID(committee []OperatorID) CommitteeID {
+	// sort
+	sort.Slice(committee, func(i, j int) bool {
+		return committee[i] < committee[j]
+	})
+	// Convert to bytes
+	bytes := make([]byte, len(committee)*4)
+	for i, v := range committee {
+		binary.LittleEndian.PutUint32(bytes[i*4:], uint32(v))
```
y0sher, line 46: operatorID is uint64, but we're putting uint32 here.. whats the correct type?

GalRogozinski: fixed to uint64

y0sher (approved):

moshe-blox (approved):

PR #52, opened 2024-09-18
Adapt fees to decalred effective balance

GalRogozinski:
# Description

Have users update their effective balance on the contract so correct fees can be calculated after [EIP-7251](https://eips.ethereum.org/EIPS/eip-7251) is deployed on Ethereum.

### DIFF
--- a/sips/adapt_fees_to_declared_effective_balance.md
+++ b/sips/adapt_fees_to_declared_effective_balance.md
@@ -1,0 +1,166 @@
+| Author | Title          | Category                             | Status  | Date               |
+|-------|----------------|--------------------------------------|---------|--------------------|
+| Gal Rogozinski & Marco Tabasco | Adapt fees to declared effective balance | Contracts & Core | open-for-discussions  |18.09.2024 |
+
+
+## Summary
+
+Creates a mechanism to incentivize users to record the correct balance of their validator with the contract.
+This balance will be used for operator and network fee calculations.
+
+## Motivation
+
+Operators on SSV Network charge validators a fee based on an expected workload. The expected workload is proportional to the the effective balance a validator has. Currently the effective balance is assumed to be a constant 32 ETH by the contract. With the introduction of [EIP-7251](https://eips.ethereum.org/EIPS/eip-7251) the effective balance of a validator could be as high as 2048 ETH.
+
+The same reasoning applies for the network fee charged to each validator registered.
+
+
+## Rationale
+
+Do not perform duties if the validator's effective balance is greater by more than a deviation configured by the contract's owner. This will incentivize validators to always register the correct amount on the contract. Effective balance may grow due to compounding rewards and it is the user's responsibility to update their effective balance before their operators pause. Paused operators will resume when the effective balance is updated on the contract.
+
+## Specification
+
+### Contract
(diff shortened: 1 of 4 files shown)


sips/adapt_fees_to_declared_effective_balance.md
```diff
@@ -0,0 +1,166 @@
...
+```solidity
+// Function to update the maximum deviation allowed for the ETH 
+// balance declared by validator owners
+// @dev Using 10000 to represent 2 digit precision: 100% = 10000, 10% = 1000
+// @dev Only the owner of the contract can perform the update
+function updateAllowedBalanceDeviation(uint64 percentage) external onlyOwner
+```
+
+```solidity
+// Function to get the maximum ETH balance deviation
```
alonmuroch, line 84: does it really need to be stored on chain? or can nodes simply parse it from events?

sips/adapt_fees_to_declared_effective_balance.md
```diff
@@ -0,0 +1,166 @@
+| Author | Title          | Category                             | Status  | Date               |
+|-------|----------------|--------------------------------------|---------|--------------------|
+| Gal Rogozinski & Marco Tabasco | Adapt fees to declared effective balance | Contracts & Core | open-for-discussions  |18.09.2024 |
+
+
```
alonmuroch, line 5: what bothers me with this approach is that a user having thousands of validators will need to update each one independently .. 

Post pectra all validators start to compound above 32 automatically? if so it will be a nightmare from UX POV

alonmuroch (changes requested):

GalRogozinski:
Conclusion from meeting:
1. Maybe the operators can act as oracles because they are incentivized to... but we may have a cheating problem. @MatusKysel reminds that if balance decreases there is no incentive to update it. @GalRogozinski say If balance decreases the user has an incentive though. @moshe-blox reminds that the operators will have to create transactions... might be an over demanding task.
2. @moshe-blox claims that the UX issues we present now are similar to the issues with topping up SSV balance.

sips/adapt_fees_to_declared_effective_balance.md
GalRogozinski: events sounds good to me

@mtabasco wants to comment?

sips/adapt_fees_to_declared_effective_balance.md
GalRogozinski: Could be also that updating clusters instead of validators will simplify things

sips/adapt_fees_to_declared_effective_balance.md
mtabasco: Yeah, if there is no major UX issue removing `getAllowedBalanceDeviation()`, the ETH balance deviation can be emitted.

sips/adapt_fees_to_declared_effective_balance.md
```diff
@@ -0,0 +1,198 @@
...
+## Summary
+
+Creates a mechanism to incentivize users to record the correct balance of their validator with the contract.
+This balance will be used for operator and network fee calculations.
+
+## Motivation
+
+Operators on SSV Network charge validators a fee based on an expected workload. The expected workload is proportional to the the effective balance a validator has. Currently the effective balance is assumed to be a constant 32 ETH by the contract. With the introduction of [EIP-7251](https://eips.ethereum.org/EIPS/eip-7251) the effective balance of a validator could be as high as 2048 ETH.
```
GalRogozinski, line 13: The workload won't increase linearly in proportion to eff balance... so maybe this should be rephrased

PR #53, opened 2024-11-28
Validators per operator cap: 500 to 1000

MatheusFranco99:
## Overview

This SIP proposes changing the validators per operator cap from 500 to 1000. This change is possible due to the recent optimizations and allows a higher profitability per operator.

### DIFF
--- a/sips/validators_per_operator_1k_cap.md
+++ b/sips/validators_per_operator_1k_cap.md
@@ -39,7 +39,7 @@ Letting $v$ as the number of validators per operator, to account for attestation
 
 $$v + max(512, v)$$
 
-Before we had $500 + max(512,500) = 1000$, now, we have $1000 + max(512,1000) = 1512$.
+Before we had $500 + min(512,500) = 1000$, now, we have $1000 + min(512,1000) = 1512$.
 
 Thus, we need to change the `PartialSignatureMessages` structure in the following way:
 
--- a/sips/validators_per_operator_1k_cap.md
+++ b/sips/validators_per_operator_1k_cap.md
@@ -1,0 +1,61 @@
+|     Author     |               Title               | Category  |       Status        |    Date    |
+| -------------- | --------------------------------- | --------- | ------------------- | ---------- |
+| Matheus Franco | 1K cap on validators per operator | Contracts | open-for-discussion | 2024-11-28 |
+
+## Summary
+
+This SIP proposes increasing the cap on the number of validators per operator from 500 to 1000.
+
+## Motivation
+
+The recent optimizations allowed operators to reduce their operational costs while running the same number of validators, thus keeping the same profits. Therefore, this allows operators to further increase the size of their validator sets, keeping a sustainable processing cost and increasing their profitability.
+
+## Performance trade-offs
+
+To understand the performance impact, we compared two settings with the same number of validators (namely, 104k), one with a cap of 500 and the other of 1000 validators per operator. Using the 500 case as a benchmark, we got the following results for the 1000 case:
+
+| Metric                             | Impact     |
+|------------------------------------|------------|
+| Maximum operator cryptography cost | 20% higher |
+| Average operator cryptography cost | 80% higher |
+| Maximum operator message rate      | 30% lower  |
+| Average operator message rate      | 22% lower  |
+
+This behavior is expected. More validators per operator reduce the number of required operators and allow Committee Consensus to aggregate more duties. This promotes a reduction in the number of exchanged messages and non-committee message ratios. On the other hand, since the partial signature phase is the heaviest influence on the cryptography cost, increasing the cap from 500 to 1k inevitably increases the cryptography cost.
+
+> [!NOTE]
+> With the Pectra fork, dropping the `CommitteeIndex` field in the attestation beacon object will allow an operator to verify all post-consensus signatures in constant time. In other terms, currently, due to different `CommitteeIndex`s, the complexity of signatures verification on the post-consensus phase is $\mathcal{O}(d)$, where $d$ is the number of duties. After the Pectra fork, it will become $\mathcal{O}(1)$, significantly reducing the overall cryptography cost.
+
+## Security
+
+Post-consensus messages may increase their average and maximum sizes as shown in the [spec changes](#partialsignaturemessages). Nonetheless, it's still way smaller than a consensus proposal message. Thus, the proposed change doesn't introduce extra concerns on network security.
+
+## Spec changes
+
+### PartialSignatureMessages
+
+The proposed change influences the number of signatures a post-consensus message may have.
+Letting $v$ as the number of validators per operator, to account for attestation and sync committee duties, the maximum number of signatures per message is computed as
+
+$$v + max(512, v)$$
+
+Before we had $500 + max(512,500) = 1000$, now, we have $1000 + max(512,1000) = 1512$.
+
+Thus, we need to change the `PartialSignatureMessages` structure in the following way:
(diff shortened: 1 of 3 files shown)


sips/validators_per_operator_1k_cap.md
```diff
@@ -0,0 +1,61 @@
...
+## Spec changes
+
+### PartialSignatureMessages
+
+The proposed change influences the number of signatures a post-consensus message may have.
+Letting $v$ as the number of validators per operator, to account for attestation and sync committee duties, the maximum number of signatures per message is computed as
+
+$$v + max(512, v)$$
+
+Before we had $500 + max(512,500) = 1000$, now, we have $1000 + max(512,1000) = 1512$.
```
GalRogozinski, line 42: seems like max should be min?

GalRogozinski:
```suggestion
Before we had $500 + min(512,500) = 1000$, now, we have $1000 + min(512,1000) = 1512$.
```

sips/validators_per_operator_1k_cap.md
```diff
@@ -0,0 +1,61 @@
...
+```
+
+### Message Validation
+
+In message validation, we need to update the following:
+- for the rule on the number of signatures contained in a post-consensus message, the limit must be changed from 1000 to 1512.
+- for the rules on the maximum byte size of a message, we have the new maximum sizes:
+  - maximum encoded `PartialSignatureMessages`: 144020 $\to$ 217748
+  - maximum encoded `SSVMessage` with encoded `PartialSignatureMessages`: 144088 $\to$ 217816
+  - maximum encoded `SignedSSVMessage` with encoded `PartialSignatureMessages`: 147588 $\to$ 221316
```
GalRogozinski, line 61: this came from our tests right?

sips/validators_per_operator_1k_cap.md
```diff
@@ -0,0 +1,61 @@
...
+|------------------------------------|------------|
+| Maximum operator cryptography cost | 20% higher |
+| Average operator cryptography cost | 80% higher |
+| Maximum operator message rate      | 30% lower  |
+| Average operator message rate      | 22% lower  |
+
+This behavior is expected. More validators per operator reduce the number of required operators and allow Committee Consensus to aggregate more duties. This promotes a reduction in the number of exchanged messages and non-committee message ratios. On the other hand, since the partial signature phase is the heaviest influence on the cryptography cost, increasing the cap from 500 to 1k inevitably increases the cryptography cost.
+
+> [!NOTE]
+> With the Pectra fork, dropping the `CommitteeIndex` field in the attestation beacon object will allow an operator to verify all post-consensus signatures in constant time. In other terms, currently, due to different `CommitteeIndex`s, the complexity of signatures verification on the post-consensus phase is $\mathcal{O}(d)$, where $d$ is the number of duties. After the Pectra fork, it will become $\mathcal{O}(1)$, significantly reducing the overall cryptography cost.
```
GalRogozinski, line 27: I think we already can and should do the optimization where CommitteeIndex is the same

sips/validators_per_operator_1k_cap.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Security
+
+Post-consensus messages may increase their average and maximum sizes as shown in the [spec changes](#partialsignaturemessages). Nonetheless, it's still way smaller than a consensus proposal message. Thus, the proposed change doesn't introduce extra concerns on network security.
```
GalRogozinski, line 31: Is this security? I think performance

GalRogozinski (changes requested):

sips/validators_per_operator_1k_cap.md
GalRogozinski: in a separate Spec PR than what you have now

GalRogozinski (commented):
Something that occurred to me that we can research...
Instead of sending one large message we can send several small sub-messages

This may be good for reducing latency because processing can start before the big message is downloaded

No need to do this optimization now

sips/validators_per_operator_1k_cap.md
```diff
@@ -0,0 +1,61 @@
...
+The recent optimizations allowed operators to reduce their operational costs while running the same number of validators, thus keeping the same profits. Therefore, this allows operators to further increase the size of their validator sets, keeping a sustainable processing cost and increasing their profitability.
+
+## Performance trade-offs
+
+To understand the performance impact, we compared two settings with the same number of validators (namely, 104k), one with a cap of 500 and the other of 1000 validators per operator. Using the 500 case as a benchmark, we got the following results for the 1000 case:
+
+| Metric                             | Impact     |
+|------------------------------------|------------|
+| Maximum operator cryptography cost | 20% higher |
+| Average operator cryptography cost | 80% higher |
```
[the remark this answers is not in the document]
MatheusFranco99: I'll add the following explanation in the text:
- if we didn't have Committee Consensus, the average would be almost 100% higher (note that non-committee cryptography has less relevance here). But committee consensus optimizes this to 80% on average.
- Why not better than 80%? When you have few validators, committee consensus doesn't have much impact. So, for the average operator, it's not that significant.
- The maximum case is the case of an operator that had 500 validators in the worst topic, and now it has 1000. The effects of committee consensus are boosted because the number of consensus duties approaches a fixed number of 32 per epoch, even though you doubled the number of validators.
Let me know if it makes sense for you? Btw, these are the numbers from the simulations. Above is my interpretation of it, not the ground truth.

sips/validators_per_operator_1k_cap.md
MatheusFranco99: Yes thanks!

sips/validators_per_operator_1k_cap.md
MatheusFranco99: Yes

sips/validators_per_operator_1k_cap.md
MatheusFranco99: Sure
Some time ago I created a task for it on Monday with the reasoning (it's on the backlog I suppose)

sips/validators_per_operator_1k_cap.md
MatheusFranco99: This is security in the sense of DoS by large messages

sips/validators_per_operator_1k_cap.md
```diff
@@ -0,0 +1,68 @@
...
+Post-consensus messages may increase their average and maximum sizes as shown in the [spec changes](#partialsignaturemessages). Nonetheless, it's still way smaller than a consensus proposal message. Thus, the proposed change doesn't introduce extra concerns on network security.
+
+## Spec changes
+
+### PartialSignatureMessages
+
+The proposed change influences the number of signatures a post-consensus message may have.
+Letting $v$ as the number of validators per operator, to account for attestation and sync committee duties, the maximum number of signatures per message is computed as
+
+$$v + max(512, v)$$
```
GalRogozinski, line 47:
```suggestion
$$v + min(512, v)$$
```

Here as well :-)

GalRogozinski (approved):
Great!

moshe-blox (approved):
LGTM 🤝

GalRogozinski:
@MatheusFranco99 
can you give me access to your branch 
OR
quickly change the status of the SIP to `spec-merged` and add it to `all.md` and `core.md`

MatheusFranco99:
@GalRogozinski 
Added you to my branch and also updated the status and added it to `all.md` and `core.md`

PR #55, opened 2025-03-12
Preconfirmation

alan-ssvlabs:


GalRogozinski (commented):
- [ ] This is for commit-boost not preconfs
- [ ] Double check commit-boost api
- [ ] Check whether commit-boost have a limit of the number of commits

Talk to commit-boost guys about:
1. checking our assumptions
2. adding endpoint for posting/fetching proofs, and tx hashes
3. Can we rely soley on proxy keys?
4. Slashing issues... commit boost protects us from contradicting commitments

GalRogozinski:
In order to continue to working on the SIP:
Divide it into 2 sections:
1. Vanilla commit-boost
2. Preconfirmation support via commit-boost

For both we need to add a Message Validation section.
For (1) currently just a decent rate limit
For (2) more complicated rules

GalRogozinski:
If we add any arbitrary rate limit, until we have simulation or real life data, we should just have a placeholder for the limit

PR #61, opened 2025-05-13
Aggregator Committee Duty

### LINKED ISSUE #56: Utilize Committee Runner for Aggregator Duties
## Description

Currently we utilize committee runner only for attestation and sync committee duties. We can still get relieve resources for aggregator duties.
We should mention how many resources we expect to save.

MatheusFranco99:
## Overview

This SIP proposes merging the Aggregator and Sync Committee Contributor duties into a single duty for the whole committee, similar to what has been done with Attestations and Sync Committee duties.

Closes #56

### DIFF
--- a/sips/aggregator_committee_consensus.md
+++ b/sips/aggregator_committee_consensus.md
@@ -189,10 +189,10 @@ const (
 	RoleCommittee                       = RunnerRole(0)
 	// RoleAggregator                   (1) deprecated
 	RoleProposer                        = RunnerRole(2)
-    // RoleSyncCommitteeContribution    (3) deprecated
+	// RoleSyncCommitteeContribution    (3) deprecated
 	RoleValidatorRegistration           = RunnerRole(4)
 	RoleVoluntaryExit                   = RunnerRole(5)
-    RoleAggregatorCommittee             = RunnerRole(6) // New
+	RoleAggregatorCommittee             = RunnerRole(6) // New
 
 	RoleUnknown = RunnerRole(-1)
 )


GalRogozinski (changes requested):
Good job!

The new domain type is for the fork that can contain many other SIPs. Even though this SIP should enter the next fork we should write SIP in a way that they can be delayed.

Also I feel the way we aggregate message is too implicit here. Let's make it more clear

sips/aggregator_committee_consensus.md
```diff
@@ -0,0 +1,275 @@
...
+
+## Improvement
+
+We evaluated the performance of the proposed change using a Monte Carlo simulation with a single committee with 1k validators, assigning beacon duties over 100 epochs.
+
+| Metric                                 | Current | New  | New/Current% |
+|----------------------------------------|---------|------|--------------|
+| **Messages/s**                         | 13.15   | 2.39 | 18%          |
+| **Bandwidth (KB/s)**                   | 8.55    | 4.56 | 53%          |
+| **Average size Per Message (KB/msgs)** | 0.65    | 1.90 | 292%         |
```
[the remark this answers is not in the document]
MatheusFranco99: Agree. Updating

MatheusFranco99:
@olegshmuelov 
Regarding this [comment](https://github.com/ssvlabs/SIPs/pull/61#issuecomment-3044759826), a new message structure is not strictly necessary. For example, this case is similar to the post-consensus signatures for attestation and sync committee duties, which are all encapsulated into the current `PartialSignatureMessages` structure. It works because the pair (`ValidatorIndex`, `SigningRoot`) uniquely identifies a (validator, beacon role, beacon committee) tuple, as I said in the doc
But I agree that it may be better to specify this structure more clearly for these cases, as this is already a fork.
What do you guys think?
@GalRogozinski @alan-ssvlabs @moshe-blox @y0sher @MatusKysel

sips/aggregator_committee_consensus.md
```diff
@@ -0,0 +1,260 @@
...
+}
+```
+
+## P2P
+
+### Message Validation
+
+- `SSVMessage.MsgID` must include a CommitteeID encoded with a 16-byte 0x00 prefix to match `ValidatorPublicKey` length. If the CommitteeID doesn't exist in the current network, it should ignore the message.
+- If a `ValidatorIndex` in `SignedPartialSignatureMessage.Message.Messages` is incorrect, considering the `ValidatorPublicKey` or the `CommitteeID` in the `MessageID`, it should ignore the message.
+- If the same `ValidatorIndex` appears more than 2 times in a `PartialSignatureMessages`, the message is rejected.
```
GalRogozinski, line 260: are these new rules?
If not make it clear

sips/aggregator_committee_consensus.md
```diff
@@ -223,6 +229,67 @@ type PartialSignatureMessage struct {
...
+> In this case, the operator $O_i$ will only possibly select validators $V_i \cap (\bigcup_{j\neq i}V_j)$.
+> - Suppose byzantine operators send a valid signature only for $v\in V_e$ and invalid ones for all other validators.
+> Once a quorum is reached, honest operators may be restricted to advance to consensus only with the duty for $v$.
+> Note that in case it receives an all-honest quorum, it may still be able to advance to consensus with more validators than just $v$.
+> 
+> While this solution prioritizes liveness over completeness,
+> it's acceptable for now as the worst-case scenarios are marginally expected.
+> Still, our context may be formally linked to the *Interactive Consistency*
+> problem, presented in the [*Reaching Agreement in the Presence of Faults*](https://lamport.azurewebsites.net/pubs/reaching.pdf)
+> paper, and more robust solutions may be explored in the future ([example](https://cgi.di.uoa.gr/~mema/publications/ic-extended.pdf)).
```
GalRogozinski: good!
you can farther downplay this problem by explaining that it is hard to exploit, since the attacker must know ahead of time what validators it holds can be selected

sips/aggregator_committee_consensus.md
```diff
@@ -239,22 +306,32 @@ type AggregatorConsensusData struct {
...
+
+
 ## P2P
 
 ### Message Validation
 
 - `SSVMessage.MsgID` must include a CommitteeID encoded with a 16-byte 0x00 prefix to match `ValidatorPublicKey` length. If the CommitteeID doesn't exist in the current network, it should ignore the message.
 - If a `ValidatorIndex` in `SignedPartialSignatureMessage.Message.Messages` is incorrect, considering the `ValidatorPublicKey` or the `CommitteeID` in the `MessageID`, it should ignore the message.
-- If the same `ValidatorIndex` appears more than 2 times in a `PartialSignatureMessages`, the message is rejected.
+- If the same `ValidatorIndex` appears more than 5 times in a `PartialSignatureMessages`, the message is rejected.
```
GalRogozinski: good fix!

sips/aggregator_committee_consensus.md
```diff
@@ -0,0 +1,338 @@
...
+
+### MessageID
+
+Messages associated to the `AggregatorRunner` and `SyncCommitteeAggregatorRunner` used to include a `ValidatorPublicKey` in `MessageID`.
+On the other hand, `AggregatorCommitteeRunner` sends messages with respect to the committee of operators and, thus, it will use `CommitteeID` in `MessageID`.
+Because `CommitteeID` is 32 bytes long, it's encoded with a 16-byte `0x00` prefix to make it the same length as a `ValidatorPublicKey` would have
+
+#### `PartialSignatureMessages`
+
+The current structures will remain unchanged as the pair (`ValidatorIndex`,`SigningRoot`) uniquely identifies a (validator, beacon role, beacon committee) tuple due to the non-collision property of the hashed signing root.
```
y0sher: I think this isn't 100% true, since you can't determine the role only from index+root. this is correct only when you assume all operators are able to calculate the same root by themselves and keep a mapping between root and role. which I think is worth mentioning.

sips/aggregator_committee_consensus.md
```diff
@@ -0,0 +1,338 @@
...
+the associated duty is submitted to the beacon chain.
+
+
+## P2P
+
+### Message Validation
+
+- `SSVMessage.MsgID` must include a CommitteeID encoded with a 16-byte 0x00 prefix to match `ValidatorPublicKey` length. If the CommitteeID doesn't exist in the current network, it should ignore the message.
+- If a `ValidatorIndex` in `SignedPartialSignatureMessage.Message.Messages` is incorrect, considering the `ValidatorPublicKey` or the `CommitteeID` in the `MessageID`, it should ignore the message.
+- For the Aggregator Committee duty, if the same `ValidatorIndex` appears more than 5 times in a `PartialSignatureMessages` (both for `PostConsensusPartialSig` and `AggregatorCommitteePartialSig`), the message is rejected.
```
y0sher: I think we should expand here about how 5 was chosen.

sips/aggregator_committee_consensus.md
```diff
@@ -0,0 +1,338 @@
...
+class AggregateAndProof(Container):
+    aggregator_index: ValidatorIndex
+    aggregate: Attestation
+    selection_proof: BLSSignature
+
+# Before Electra
+class Attestation(Container):
+    aggregation_bits: Bitlist[MAX_VALIDATORS_PER_COMMITTEE]
+    data: AttestationData
+    signature: BLSSignature
```
GalRogozinski: wasn't there `committeeIndex`?
regardless this isn't needed anymore I guess

sips/aggregator_committee_consensus.md
GalRogozinski: I also forget, so good idea 😆

sips/aggregator_committee_consensus.md
MatheusFranco99: Adding to make it more explicit.

sips/aggregator_committee_consensus.md
MatheusFranco99: You are correct, but this is inside `AttestationData` (not `Attestation` itself),
```go
type AttestationData struct {
	Slot            Slot
	Index           CommitteeIndex
	BeaconBlockRoot Root `ssz-size:"32"`
	Source          *Checkpoint
	Target          *Checkpoint
}
```
which was set to `0` after the Electra fork.

GalRogozinski (approved):
Great!

sips/aggregator_committee_consensus.md
```diff
@@ -0,0 +1,344 @@
...
+const (
+	RoleCommittee RunnerRole = iota
+	RoleAggregatorCommittee
+	RoleProposer
+
+	RoleValidatorRegistration
+	RoleVoluntaryExit
+
+	RoleUnknown = -1
+)
```
GalRogozinski: @MatheusFranco99 we talked about changing this as well, right?

GalRogozinski (approved):
Awesome

sips/aggregator_committee_consensus.md
```diff
@@ -0,0 +1,347 @@
...
+
+type AggregatorConsensusData struct {
+	Version spec.DataVersion // Beacon version
+
+	// Aggregator duties
+	Aggregators     []AssignedAggregator
+    // unique list of the existing beacon committees, i.e., a subset of the [0,...,63] list
+	AggregatorsCommitteeIndexes  []uint64
+	// list of ssz encoded phase0.Attestation or electra.Attestation (depending on the version), one for each beacon committee index
+	AggregatedAttestations     	 [][]byte
```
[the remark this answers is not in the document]
GalRogozinski: @MatheusFranco99 
Reminder to fix this in the meanwhile

PR #64, opened 2025-05-24
Merge Post Consensus Messages into QBFT Commit Messages

alan-ssvlabs:


MatheusFranco99:
I think I have written a proof somewhere that this is unsafe and that a malicious user can slash the validator.

alan-ssvlabs:
@GalRogozinski @MatheusFranco99 
But if the consensus has reached to commit phase before a round change, the round change message should include the previous prepared justification and the new round should use the same proposal as prepared in the last round. ([reference](https://entethalliance.org/specs/qbft/v1/#:~:text=if%20%20hasReceivedProposalJustificationForLeadingRound(current))) So no new value will be signed.

GalRogozinski:
This is still not safe theoretically because you may have an honest node that won't sign the decided value. If there is a malicious actor, we won't reach consensus.

There can even be liveness issues with slashing protection w.o. a byzantine actor

PR #67, opened 2025-05-28
Finality based event syncing 

kchojn:


dknopik (commented):
Looks good!

y0sher:
> IMO, we can introduce some mechanism to mitigate the long non-finality issue a bit. Basically, we want to be able to remove validators in order to run them somewhere else, or at least exit them from the active validator set.
> ### Always perform exit events
> 
> Process validator exit events regardless of finality. This is safe, since it does not change validator ownership: We do not have a slashing risk as the validator is not removed from the cluster (and therefore can not be readded to another). Also, it does not matter if the event is reorged, as it does not affect the state.
> ### Pending removal flag
> 
> Another idea is immediately checking all new blocks regardless of finality. For all validator removals in this block, the validator is marked "pending removal". For all "pending removal" validators:
> 
>     * Do not perform duties
> 
>     * Do not forward messages where one of the following is true:
>       
>       * Message ID contains the validator
>       * The message is a partial signature message and one of the contained partial signatures is for the validator
> 
> 
> The only way to remove a "pending removal" flag is by actually removing the validator (i.e. finalizing a block with a remove validator event). This should be reasonably safe: The user can observe if the transaction causes the validator to stop. Even if the event reorgs, the operators will remember the flag, and not perform duties. Great care should be taken to persist this flag! Ideally, the node should remember it even if the user decides to resync the node, e.g. by storing this information in a separate file.
> 
> The disadvantage of this is that the user still can not really use the validator differently, as the user does not have a slashing protection database for that validator. However, it is not impossible to e.g. get the slashing protection data from an operator and use it in an conventional validator client.
> 
> Note that it is not possible to make this 100% safe, as operators may lose this information, and therefore is not my favoured solution.
> ### Disable validator config
> 
> This is similar to above solution, except it is just support for disabling a certain validator in an operator via a configuration file. The operator configure do this at the request of the validator owner. This would be very helpful for private clusters.
> 
> Note that the paragraph regarding the slashing protection data from above still applies.
> 
> Happy to hear comments on these ideas from you!

Hi @dknopik ,
All of this ideas sound good to me as we were considering the exit/removal issue and it's sounds reasonable to mitigate this way, achieving the robustness against re-orgs but not hurting the UX or removing validators. Thanks for taking the time to review and add your input. We will update this PR and code soon to reflect the changes.

I'm not sure the `disable validator config` feature has to be part of this though, I understand why it's a feature that might be desired but doesn't have to do much with syncing.

GalRogozinski:
I think we said this isn't a fork but I am not sure all of a sudden.
This could cause issues with Message Validation

PR #90, opened 2026-04-10
Add Security Considerations section to the SIP template

oleg-ssvlabs:
Closes https://github.com/ssvlabs/ssv-node-board/issues/517

PR #98, opened 2026-06-24
Add SIP authoring guidance

GalRogozinski:
## Summary

Adds repository-level agent guidance for writing SIPs:

- `AGENTS.md` points agents to `CLAUDE.md`.
- `CLAUDE.md` explains how to draft precise, concise SIPs using the repo template and RFC-style requirement language.

The guidance emphasizes enough specificity for independent implementations in different languages to remain compatible.

## Validation

- Reviewed the existing SIP template and SIP lifecycle document.
- Renamed the agent guidance files to the canonical `AGENTS.md` and `CLAUDE.md` casing for case-sensitive filesystems.
- Updated the `AGENTS.md` link to target `CLAUDE.md`.
- Ran `git diff --cached --check` before committing.


momosh-ssv (commented):
Content looks accurate against `template_sip.md` and `sips/sip0.md`, and the RFC-keyword style is applied consistently. One portability point on the filenames, plus a follow-on nit on the link.

CLAUDE.md
```diff
@@ -0,0 +1,51 @@
+# SIP Authoring Guidance
```
momosh-ssv, line 1: Should we consider naming these `CLAUDE.md` and `AGENTS.md` (exact name, lowercase `.md`)? Claude Code looks for the exact name `CLAUDE.md`, and `AGENTS.md` is the cross-tool standard. 

On case-sensitive filesystems (Linux/CI), `Claude.MD` won't be auto-discovered, so this guidance would silently not apply there — it only loads on macOS/Windows by case-insensitivity, which is the whole purpose of the file. Fix direction: rename both files to the canonical casing.

Agents.MD
```diff
@@ -0,0 +1 @@
+[Claude.MD](./Claude.MD)
```
momosh-ssv: Worth considering: this link resolves only because the target filename casing matches. If we rename to `CLAUDE.md` per the other comment, updating this to `[CLAUDE.md](./CLAUDE.md)` keeps them in sync and avoids a broken link on case-sensitive systems.

CLAUDE.md
GalRogozinski: Fixed in 2c6ff19: renamed `Claude.MD` to canonical `CLAUDE.md` and `Agents.MD` to `AGENTS.md` so both files are discoverable on case-sensitive filesystems.

Agents.MD
GalRogozinski: Fixed in 2c6ff19: updated the `AGENTS.md` link to `[CLAUDE.md](./CLAUDE.md)` so the link target matches the canonical filename casing.

MatheusFranco99 (dismissed):
lgtm

momosh-ssv (dismissed):
nice, approved!

CLAUDE.md
```diff
@@ -0,0 +1,51 @@
...
+relevant table of contents when the SIP is ready for review.
+
+Each SIP SHOULD include:
+
+- metadata table: author, title, category, status, dependencies, and date;
+- summary: the problem and proposed change in a few sentences;
+- rationale and design goals: why this change is needed and what constraints it
+  optimizes for;
+- specification: normative behavior, formats, algorithms, validation rules, and
+  test expectations.
```
shane-moore: The template has a fourth body section this list omits: **Security Considerations**. The template requires it even when there's no security impact ("If the SIP has no security impact, state so explicitly and explain why"), so guidance following this list would produce SIPs missing a mandated section. Suggest adding a bullet:

```
- security considerations: new attack surfaces, trust assumptions, and
  interactions with slashing, consensus safety, and liveness; state
  explicitly if there is no security impact.
```

GalRogozinski: Fixed in 0f522e1: added the `security considerations` bullet to the required-sections list, and merged `main` to pick up the template's Security Considerations section (#90).

shane-moore (approved):

MatheusFranco99 (approved):

momosh-ssv (approved):

PR #94, opened 2026-05-06
ePBS (EIP-7732) Support

shane-moore:
This SIP describes the ssv spec changes needed to keep operators performing validator duties correctly after ePBS (EIP-7732) lands in the consensus layer Gloas fork. Covers earlier slot deadlines, `AttestationData.Index` propagation through `BeaconVote`, the new PTC committee duty, the `produceBlockV4` proposer flow (self-build vs external-builder variants), and the new `SignedProposerPreferences` broadcast.

PR #101, opened 2026-07-23
Add early RANDAO pre-consensus emission SIP

shane-moore:
Operators may emit the existing Proposer-duty RANDAO partial signature up to 2 slots before the proposal slot, so clusters can reconstruct `randao_reveal` before the slot starts instead of inside the post-Gloas ~3s block-production budget.

No new duty, role, message kind, domain, topic, or container is introduced. Protocol changes are limited to target-slot activation, message-validation timing, message-kind-scoped ordering, and complete-schedule duty handling. Receivers MAY additionally retain an otherwise-valid Unknown-duty candidate locally, but retention is optional and does not change the initial IGNORE gossip verdict.

The rules activate from the stamped proposal slot `S`: they apply when `epoch(S) >= GLOAS_FORK_EPOCH`, including when the permitted window for the first Gloas slots begins before the fork in wall-clock time. The SSV fork schedule is not an input.

Rationale and design discussion: https://github.com/ssvlabs/SIPs/discussions/100
Addresses: https://github.com/ssvlabs/ssv-spec/issues/373


sips/early_randao_preconsensus.md
```diff
@@ -0,0 +1,115 @@
+| Author | Title | Category | Status | Dependency SIP | Date |
+| ------ | ----- | -------- | ------ | -------------- | ---- |
+| Shane Moore | Early RANDAO Pre-Consensus Emission | Core | draft | ePBS (EIP-7732) Support | 2026-07-23 |
+
+[Discussion](https://github.com/ssvlabs/SIPs/discussions/100)
+
+**Summary**
+
+Operators may emit their existing block-proposal RANDAO partial signature up to 2 slots before the proposal slot, so the cluster reconstructs `randao_reveal` before the slot starts instead of inside the post-Gloas ~3s block-production budget. No new duty, role, message kind, domain, topic, or container: the existing Proposer-duty `RandaoPartialSig` message is emitted earlier, stamped with the proposal slot as today. Changes are confined to message-validation timing rules plus a bounded receive-side buffer. Activates at the Gloas-aligned SSV fork; the dependency on the ePBS SIP is activation coupling only.
```
iurii-ssv: The dependency "ePBS (EIP-7732) Support" is still open PR #94 (not yet in `sips/`). Suggest linking it directly here (and/or in the header row) and noting the lifecycle coupling per [SIP-0](https://github.com/ssvlabs/SIPs/blob/main/sips/sip0.md): this can't advance to `last-call` until #94 is approved, and would be rejected if #94 is. `GLOAS_FORK_EPOCH` and `ProposerPreferences` also resolve there, so the link makes those references traceable.

sips/early_randao_preconsensus.md
```diff
@@ -0,0 +1,115 @@
...
+| `MAX_QUARANTINED_MESSAGES` | 4096 |
+
+**Qualifying message**
+
+A qualifying randao partial MUST satisfy all of:
+
+- Proposer-role `MessageID`; type `RandaoPartialSig`; exactly one `PartialSignatureMessage` entry;
+- `slot` = the proposal slot `S`; signed object `SSZUint64(epoch(S))` under `DOMAIN_RANDAO`, domain epoch `epoch(S)`;
+- canonical SSZ; deterministic BLS share signature; deterministic RSA (PKCS#1 v1.5) operator signature; exactly one outer signer, equal to the embedded operator ID;
+- eligibility predicate: `fork_at_slot(S - EARLY_RANDAO_LEAD) == fork_at_slot(S)` and `epoch(S) >= GLOAS_FORK_EPOCH`.
```
iurii-ssv: `GLOAS_FORK_EPOCH` appears in this normative predicate but isn't in the Constants table or cross-referenced (it's defined in the dependency ePBS SIP — see the note on line 9). Either add it to the Constants table or cross-reference its source so the predicate is self-contained. Same for `ProposerPreferences` in Rationale.

Minor, related: `SLOT_DURATION` is only introduced as "as usual" in the notation (line 24) yet drives the Earliness formula — consider listing it in the Constants table too, since Gloas slot timing is the whole premise.

sips/early_randao_preconsensus.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+inclusive on the accept side; strictly greater is IGNORE. This replaces the generic clock tolerance for this rule only. Lateness is unchanged (existing Proposer-role TTL against `msg.slot`). All other message classes keep their existing allowances.
+
+*Slot ordering.* Qualifying randao partials are exempt from the per-(operator, `MessageID`) highest-seen-slot rule in both directions: they MUST NOT be rejected for a slot below the high-water mark, and they MUST NOT advance the high-water mark applied to other message classes on the same `MessageID` (randao shares its `MessageID` with the proposal's consensus and post-consensus traffic). The exemption applies only after the qualifying-message checks pass; earliness, lateness, the duplicate limit of 1, and canonical form bound these messages instead.
+
+*Duty assignment.* Tri-state on the receiver's proposer-duty view for `epoch(msg.slot)`:
+
+- Known (a successfully completed fetch for the epoch is cached; an empty result is Known) and assigned at exactly `msg.slot`: accept.
+- Known and not assigned: IGNORE. Never REJECT (receiver views can be stale or re-orged).
```
iurii-ssv: Retention (IGNORE-AND-RETAIN) covers only the **Unknown** state, but proposer duties are known just one epoch ahead and can shift on a re-org across the epoch N-1/N boundary — exactly the window in play when `S` is in the first `EARLY_RANDAO_LEAD` slots of an epoch (so `S - EARLY_RANDAO_LEAD` is in the prior epoch).

A receiver whose cached view predates the re-org lands on this `Known`-and-not-assigned branch → IGNORE → the seen-cache marks the message → it is *not* retained, so the honest early share is permanently dropped and recovered only by the in-slot fallback. That's acceptable, but since the spec is otherwise careful to keep honest shares out of the permanent-loss class (Rationale, line 20), please state explicitly that stale-`Known`-via-reorg is intentionally covered by the in-slot fallback rather than retention. A test vector for this path would also be worth adding to line 100.

sips/early_randao_preconsensus.md
```diff
@@ -0,0 +1,115 @@
...
+- Unknown (no successfully completed fetch cached; failures count as Unknown): IGNORE-AND-RETAIN, below.
+
+Receivers SHOULD hold next-epoch proposer duties by the tail of each epoch.
+
+**Unknown-duty retention (IGNORE-AND-RETAIN)**
+
+A message reaching the duty check in the Unknown state is retained locally before the gossip verdict IGNORE is returned: not forwarded, no penalty, but kept so the seen-cache mark cannot permanently discard it.
+
+- Admission: passed every rule except duty assignment, and `epoch(msg.slot)` is Unknown.
+- Key: (operator, validator, `msg.slot`), at most 1 entry. Global capacity `MAX_QUARANTINED_MESSAGES` per node.
```
iurii-ssv: Two asks on capacity:

1. `MAX_QUARANTINED_MESSAGES = 4096` (and the `~600 validators` attacker threshold in Security, line 113) are stated without derivation. A one-line rationale for each would help implementers size and reason about this.
2. Beyond the attacker case, please address the **legitimate large-node outage**: during a broad duty-fetch failure, honest early shares across many validators can compete for the 4096 slots and self-evict via the deterministic eviction order (line 84), silently dropping honest shares with no adversary involved. Worth a sentence on whether 4096 is expected to cover realistic large-node duty-fetch outages, or how an operator should reason about that.

sips/early_randao_preconsensus.md
```diff
@@ -0,0 +1,115 @@
...
+- Expiry: delete an entry once a newly received copy would fail the lateness rule.
+- Promotion: when the epoch's duty fetch completes, each entry is decided immediately: assigned at exactly `msg.slot` promotes, anything else deletes. A promoted message is processed exactly as a newly received accepted message, with all stateful rules applied at promotion time (a normally accepted interim copy for the same (signer, slot) makes the promoted entry the count-limit duplicate).
+- Accounting: the retention quota mutates at retention time; ordinary validation state does not mutate on the IGNORE path.
+
+Shares that are neither promoted nor normally accepted MUST NOT be fed to signature collection or any cache.
+
+**Reconstruction and consumption**
+
+Unchanged. The reconstructed reveal is a per-epoch value: an implementation MAY serve a proposal at slot `Y` from a reconstruction built via shares stamped for slot `X` of the same epoch, provided every contributing share was promoted or normally accepted. Wire messages remain per-slot-stamped.
```
iurii-ssv: This reads as conflicting with the Duty-assignment rule (line 71), which only accepts/promotes a share when the validator is "assigned at exactly `msg.slot`". A share stamped for slot `X` therefore only becomes "promoted or normally accepted" if there's a real duty at `X` — so serving a proposal at `Y` from `X`-stamped shares is only reachable when the validator is assigned at **both** `X` and `Y` in the epoch (two proposals, same epoch). If that's the intended scenario, please say so; otherwise the per-epoch reuse allowance and the exact-slot duty check appear contradictory.

iurii-ssv (dismissed):
LGTM overall, the suggestions below are mostly minor clarifications.

sips/early_randao_preconsensus.md
shane-moore: Right, and it's slightly worse than stated: the in-slot fallback doesn't re-deliver the bytes to that receiver either. The origin's in-slot re-emission is byte-identical, so its own publish-side dedup absorbs it and it never reaches the wire, and even a restarted origin's re-publish dies in the receiver's seen-cache (TTL ~6.4 min in both clients, vs a 24s window). A receiver that IGNOREd the early copy cannot obtain that share again in time at all. What actually bounds the loss is the round-change path: the stale receiver can't reconstruct, and if it leads round 1, a fresh-view peer takes over on round change.

The stale window itself is narrow in both implementations: go-ssv re-fetches proposer duties on head-event reorgs (current epoch iff `current_duty_dependent_root` changed, next epoch on any reorg, deadline slot+2), and Anchor re-polls proposer duties every slot. So a post-reorg stale Known view lasts about one to two slots with healthy BN connectivity. A concurrent duty-fetch failure extends it, but by the Known definition (a cached successful fetch persists) that correctly stays Known, so retention never sees it either way.

Extending retention to Known-not-assigned would be the alternative, but it deletes the attacker's duty-fetch-failure precondition (any operator could fill the buffer at will) and would need reorg-triggered re-evaluation of the buffer, a worse trade for a loss class already in the round-change cost bucket.

Addressed in ecf7b3c: the duty-assignment section now states the loss class explicitly, and the stale-Known reorg test vector is added.

sips/early_randao_preconsensus.md
shane-moore: Not contradictory, but underspecified, agreed. "Promoted or normally accepted" is evaluated at receipt/promotion time against the receiver's then-current duty view and never re-evaluated at consumption, which makes the allowance reachable two ways: (a) two proposals for the validator in the same epoch (your reading), and (b) a reorg that moves the proposal from X to Y after X-stamped shares were accepted under the pre-reorg view, which is the case the allowance mainly exists for: a completed early collection survives the shift, and the signed object is identical either way (`SSZUint64(epoch)`).

It's a MAY because implementations naturally split: Anchor keys signature collection by signing root (epoch-derived for randao), so it serves Y from X-stamped shares with no special-case code, while the go-ssv runner requires message slot == duty slot for pre-consensus (`validatePartialSigMsg`), so it doesn't, and both are conformant.

Addressed in 1cd0d6e: named both cases, added the point-in-time clause, and added a reuse test vector.

sips/early_randao_preconsensus.md
shane-moore: Added both. Derivations now in the text: an entry is one signed message (~0.5 KB: 96B BLS share sig + 256B RSA operator sig + envelope), so 4096 caps memory at ~2 MB. The ~600 comes from the admissible stamp span: at any instant a (operator, validator) pair has ~7 admissible slot stamps (3-slot proposer lateness TTL behind, 2-slot lead ahead, the current slot, plus tolerances), and retention only admits stamps whose epoch is Unknown at the victim, so the attacker's best case (current epoch Unknown mid-epoch) needs 4096/7 ≈ 585 shared validators; milder outages push the threshold higher.

On the honest large-node outage: honest entries additionally require a real proposer duty in the Unknown epoch, which bounds them at roughly one SSV proposal per slot network-wide, times at most 13 committee operators, times the ~7-slot span: under 100 entries, regardless of outage duration (the span slides and expiry drains it) and regardless of node size (admission isn't scoped to local validators, so demand tracks the network's proposal rate, not the node's validator count). Honest self-eviction can't happen at 4096; reaching the cap requires fabricated stamps, which is the attacker case.

Addressed in daf4c77: sizing rationale in the retention section, derivation in the security bullet, and an explicit allowance for scoping retention to committees the receiver participates in.

sips/early_randao_preconsensus.md
shane-moore: Good call. Linked #94 from the header and added the SIP-0 coupling note: this cannot advance to last-call before #94, and is rejected with it if #94 is rejected. Addressed in 4e761cf.

sips/early_randao_preconsensus.md
shane-moore: Agreed on self-containment. Cross-referenced `GLOAS_FORK_EPOCH` and `ProposerPreferences` to the ePBS SIP rather than duplicating them into this SIP's constants table (duplicating invites drift), and gave `SLOT_DURATION` an explicit definition in the notation instead of "as usual". Addressed in 7afb541.

sips/early_randao_preconsensus.md
```diff
@@ -0,0 +1,125 @@
...
+
+**Rationale & Design Goals**
+
+- No new duty: RANDAO always has an in-slot consumer (the Proposer duty), and a separate duty would still need the in-slot pre-consensus as its fallback. This differs from `ProposerPreferences` (defined in the ePBS SIP, [#94](https://github.com/ssvlabs/SIPs/pull/94)), which must reconstruct before the slot and is mutable.
+- The 2-slot window is chosen for operational margin, not gossip latency. The window is the fork-pinned receiver rule; the emission moment inside it is producer policy, tunable post-fork without a protocol change.
+- The per-(signer, slot) duplicate limit stays 1. Each logical partial has exactly one valid byte encoding, gossip message identifiers are content-derived, and gossip layers deduplicate before validation, so a second copy either never reaches validation or is the receiver's first copy.
+- Because gossip layers never unmark a seen message, an IGNORE of an early partial permanently discards that operator's share for the window. The Unknown-duty retention rule and the dedicated clock tolerance exist to keep honest shares out of that failure class.
```
iurii-ssv: Non-blocking, follow-up from the reorg thread. This now overstates the guarantee: the stale-`Known` paragraph under *Duty assignment* documents an honest early share that is permanently discarded and explicitly **not** retained, so "keep honest shares out of that failure class" is no longer unconditional.

Suggest something like "... exist to keep honest shares out of that failure class, with the stale-`Known` reorg case as the documented exception (see *Duty assignment*)." Someone reading only the Rationale would otherwise take away a guarantee the Specification does not make.

sips/early_randao_preconsensus.md
```diff
@@ -0,0 +1,125 @@
...
+- the two-direction ordering exemption with proposals at consecutive slots for one validator;
+- occupied-key two-tier duplicates and distinct-bytes REJECT;
+- the eligibility predicate at fork boundaries (first `EARLY_RANDAO_LEAD` slots of a fork epoch ineligible);
+- restart cases, late mesh join, cross-epoch stamps.
+
+Before mainnet activation guidance, testnet measurement MUST quantify early-share miss incidence at duty start (stratified by emission lead, receiver restarts, partition duration, leader round) and MUST demonstrate round-2 viability under Gloas timing, meaning block publication before the post-Gloas useful-block deadline.
+
+**Security Considerations**
+
+- Deliberately ineligible early emission only starves the attacker's own share (only the signer can produce the bytes). The clock tolerance, Unknown-state retention, and IGNORE-not-REJECT choices keep honest shares out of the permanent-loss class.
```
iurii-ssv: Same nit as in the Rationale: "keep honest shares out of the permanent-loss class" reads as absolute, but the stale-`Known` reorg path puts an honest share in exactly that class by design. Since this bullet is where a reader looks for the residual risks, a pointer to the *Duty assignment* paragraph would keep the two sections consistent.

iurii-ssv (dismissed):
Just two non-blocking nits.

sips/early_randao_preconsensus.md
shane-moore: Softened. Left it open-ended rather than naming one exception: eviction and restart-loss of retained entries are in the same class, so it points at *Duty assignment* and Security Considerations instead of enumerating.

[c3d4640](https://github.com/ssvlabs/SIPs/pull/101/commits/c3d464064cf31db1c8b148628bf69a7b93c87f71)

sips/early_randao_preconsensus.md
shane-moore: Done, with the *Duty assignment* pointer. Also noted in the restart bullet that unpromoted retained entries don't survive a restart, so the residual is stated where the cross-reference lands.

[c3d4640](https://github.com/ssvlabs/SIPs/pull/101/commits/c3d464064cf31db1c8b148628bf69a7b93c87f71)

iurii-ssv (dismissed):
LGTM!

sips/early_randao_preconsensus.md
```diff
@@ -0,0 +1,137 @@
...
+| Shane Moore | Early RANDAO Pre-Consensus Emission | Core | draft | [ePBS (EIP-7732) Support (#94)](https://github.com/ssvlabs/SIPs/pull/94) | 2026-07-23 |
+
+[Discussion](https://github.com/ssvlabs/SIPs/discussions/100)
+
+Depends on the ePBS SIP (currently [PR #94](https://github.com/ssvlabs/SIPs/pull/94)); per [SIP-0](sip0.md), this SIP cannot move to last-call before that SIP is approved, and is rejected if it is rejected (unless changed to remove the dependency).
+
+**Summary**
+
+Operators may emit their existing block-proposal RANDAO partial signature up to 2 slots before the proposal slot, so the cluster reconstructs `randao_reveal` before the slot starts instead of inside the post-Gloas ~3s block-production budget. No new duty, role, message kind, domain, topic, or container: the existing Proposer-duty `RandaoPartialSig` message is emitted earlier, stamped with the proposal slot as today. Changes are confined to message-validation timing, ordering, and duty-handling rules plus a bounded receive-side buffer. Activates at `GLOAS_FORK_EPOCH`, the same Ethereum gate the ePBS SIP uses for its own validation rules; the dependency on the ePBS SIP is activation coupling only.
```
MatheusFranco99: > up to 2 slots before the proposal slot

Hmm, just by curiosity, why 2 slots, exactly?
Also, is there an edge case, like first epoch slot, in which one can't do 2 slots or even 1 in advance?
And if so how should them behave?
Possibly, clear edge-cases are the 1st and 2nd slots of the fork, right?

sips/early_randao_preconsensus.md
```diff
@@ -0,0 +1,137 @@
...
+
+**Rationale & Design Goals**
+
+- No new duty: RANDAO always has an in-slot consumer (the Proposer duty), and a separate duty would still need the in-slot pre-consensus as its fallback. This differs from `ProposerPreferences` (defined in the ePBS SIP, [#94](https://github.com/ssvlabs/SIPs/pull/94)), which must reconstruct before the slot and is mutable.
+- The 2-slot window is chosen for operational margin, not gossip latency. The window is the fork-pinned receiver rule; the emission moment inside it is producer policy, tunable post-fork without a protocol change.
+- The per-(signer, slot) duplicate limit stays 1. Each logical partial has exactly one valid byte encoding, gossip message identifiers are content-derived, and gossip layers deduplicate before validation, so a second copy either never reaches validation or is the receiver's first copy.
```
MatheusFranco99: >  gossip layers deduplicate before validation, so a second copy either never reaches validation or is the receiver's first copy.

Just to add: byzantine duplicates with a different signature for the same (duty, signer) won't be dropped by the naive gossip hash deduplication, but it will be by our app message validation logic that accepts only 1 partial beacon signature per signer per type (pre- or post- consensus) and per duty.

sips/early_randao_preconsensus.md
```diff
@@ -0,0 +1,137 @@
...
+- Because gossip layers never unmark a seen message, an IGNORE of an early partial permanently discards that operator's share for the window. The Unknown-duty retention rule and the dedicated clock tolerance narrow that failure class for honest shares without eliminating it; the residual cases are documented in *Duty assignment* (stale-`Known` reorg), the retention admission rule (refresh-pending losses), and Security Considerations.
```
MatheusFranco99: > never unmark a seen message

I think they do so, but I also think we are just fine in here since it should be a window of 1 or 2 epochs. We just need to confirm.

sips/early_randao_preconsensus.md
```diff
@@ -0,0 +1,137 @@
...
+
+**Specification**
+
+Notation: `S` is a proposal slot; `slot_start(s)` and `epoch(s)` as usual; `SLOT_DURATION` is the network's beacon-chain seconds-per-slot (unchanged at Gloas, which alters only intra-slot timing). REJECT penalizes the delivering peer; IGNORE drops without forwarding or penalty.
+
+| Constant | Value |
+| -------- | ----- |
+| `EARLY_RANDAO_LEAD` | 2 slots |
+| `EARLY_RANDAO_CLOCK_TOLERANCE` | 1000 ms |
+| `MAX_QUARANTINED_MESSAGES` | 4096 |
```
MatheusFranco99: I think it could be useful to add a third column for a short description for each so the reading flow provides more intuition, what do you think?
For example `EARLY_RANDAO_CLOCK_TOLERANCE` appears only after some time in the document.

sips/early_randao_preconsensus.md
```diff
@@ -0,0 +1,137 @@
...
+
+**Qualifying message**
+
+A qualifying randao partial MUST satisfy all of:
+
+- Proposer-role `MessageID`; type `RandaoPartialSig`; exactly one `PartialSignatureMessage` entry;
+- `slot` = the proposal slot `S`; signed object `SSZUint64(epoch(S))` under `DOMAIN_RANDAO`, domain epoch `epoch(S)`;
+- canonical SSZ; deterministic BLS share signature; deterministic RSA (PKCS#1 v1.5) operator signature; exactly one outer signer, equal to the embedded operator ID;
+- eligibility predicate: `S >= EARLY_RANDAO_LEAD` and `epoch(S) >= GLOAS_FORK_EPOCH` (the Ethereum Gloas fork epoch, as used by the ePBS SIP, [#94](https://github.com/ssvlabs/SIPs/pull/94)).
```
MatheusFranco99, line 43: > `S >= EARLY_RANDAO_LEAD`

Hmm, I got confused here... the slot number should be above 2? Or it was more in the sense:
- First enabled slot ever: $S - \text{[GLOAS FORK EPOCH.first slot]} \geq 2$?
- Or distance to duty slot: $0 \leq S (duty.slot) - now.slot \leq 2$ ?

sips/early_randao_preconsensus.md
```diff
@@ -0,0 +1,137 @@
...
+
+The predicate is a pure function of `S`, evaluated identically by producers and receivers, never re-evaluated against wall-clock time. Activation is Ethereum-gated by `epoch(S) >= GLOAS_FORK_EPOCH`, matching the ePBS SIP's own validation gating; the `S >= EARLY_RANDAO_LEAD` conjunct keeps the producer's emission window well defined and excludes the first `EARLY_RANDAO_LEAD` slots at genesis. The SSV fork schedule plays no part, and an emission window that spans an SSV fork activation is fine: every fork-sensitive artifact of a partial signature (gossip topic, domain, role gating) is derived from the message's stamped slot rather than from the moment it was emitted, so such a message is published, validated, and consumed entirely under the fork active at `S`. The same holds for a window spanning the Ethereum fork boundary, which this predicate likewise permits. Containers violating the canonical form fall to existing structural rules (REJECT); BLS-share validity is not evaluated during message validation (see Message validation). Messages failing the predicate, and all non-randao messages, keep today's validation unchanged.
```
MatheusFranco99: > the first `EARLY_RANDAO_LEAD` slots at genesis

Ah, got it. But does it really make sense to do so? Shouldn't it be the first two of the activation epoch?

sips/early_randao_preconsensus.md
```diff
@@ -0,0 +1,137 @@
...
+
+**Producer behavior**
+
+For a locally known proposer duty at eligible slot `S`, an operator:
+
+- MAY broadcast its qualifying randao partial at any wall-clock time in `[slot_start(S - EARLY_RANDAO_LEAD), slot_start(S))`;
```
MatheusFranco99, line 55: But what about the first slot of the fork?
Is there a warm-up, or, for this case, it would end up just sending on its slot and that's it?

sips/early_randao_preconsensus.md
```diff
@@ -0,0 +1,137 @@
...
+- MUST NOT broadcast it before `slot_start(S - EARLY_RANDAO_LEAD)` by its own clock;
+- SHOULD delay emission at least 500 ms past `slot_start(S - EARLY_RANDAO_LEAD)` (assumed maximum pairwise honest clock disparity: 1 s);
+- SHOULD still execute the existing in-slot emission at `S` unconditionally; a running origin's identical re-publish is absorbed by its own gossip layer (expected, not an error), and a restarted origin's re-publish aids recovery;
```
MatheusFranco99: Wait, so it always also try to emit again once the slot starts?
I think there could be a problem here since re-creating a message and BLS-signing it takes time. Only if we have it cached...
But maybe a flag about the message being published well or not is enough, no? What do you think?

sips/early_randao_preconsensus.md
```diff
@@ -0,0 +1,137 @@
...
+
+**Producer behavior**
+
+For a locally known proposer duty at eligible slot `S`, an operator:
+
+- MAY broadcast its qualifying randao partial at any wall-clock time in `[slot_start(S - EARLY_RANDAO_LEAD), slot_start(S))`;
+- MUST NOT broadcast it before `slot_start(S - EARLY_RANDAO_LEAD)` by its own clock;
+- SHOULD delay emission at least 500 ms past `slot_start(S - EARLY_RANDAO_LEAD)` (assumed maximum pairwise honest clock disparity: 1 s);
```
MatheusFranco99: Hmm, sorry I got lost in here. What is the purpose of doing that? Trying to let the receiver accept the msg due to the possibility of it having an earlier clock time and still being in the duty slot - 3? We can align msg validation for that rather than the logic (as we already do if I'm not mistaken)

MatheusFranco99 (commented):
Good work @shane-moore !!
First review batch :)

sips/early_randao_preconsensus.md
iurii-ssv, line 43: The predicate is a pure function of `S` with no earliness term, so per line 65 *every* post-Gloas randao partial is a candidate — ordinary in-slot ones included, and the line-56 re-emission. The *Duty assignment* tri-state therefore governs all post-Gloas randao, not just early copies.

That changes an existing rule: go-ssv accepts and forwards a randao whose epoch view is unfetched at the first slot of an epoch ([`common_checks.go#L135`](https://github.com/ssvlabs/ssv/blob/a1bcd468ff667d88c7affa58dedb3345e3467f96/message/validation/common_checks.go#L135) → [`ValidationAccept`](https://github.com/ssvlabs/ssv/blob/a1bcd468ff667d88c7affa58dedb3345e3467f96/message/validation/errors.go#L189)); under this SIP it's Unknown → IGNORE ([`ErrNoDuty` has no `reject` flag → `ValidationIgnore`](https://github.com/ssvlabs/ssv/blob/a1bcd468ff667d88c7affa58dedb3345e3467f96/message/validation/errors.go#L171)) → not forwarded. Narrow in practice — only the Unknown branch differs, and per line 85 that's a beacon-node outage — but it's a behavior change the Summary's "confined to" understates.

Ask: say the tri-state replaces existing randao duty handling for all post-Gloas randao, and add a vector for an in-slot randao under an Unknown view.

Separately: candidates are type-scoped, but go-ssv's earliness check is keyed by *role* ([`validateSlotTime`](https://github.com/ssvlabs/ssv/blob/a1bcd468ff667d88c7affa58dedb3345e3467f96/message/validation/common_checks.go#L16), [`earlySlotAllowance(role)`](https://github.com/ssvlabs/ssv/blob/feeccd8f323486a67b6f4cf367b9c7549c10c048/message/validation/common_checks.go#L52)). At role granularity the 2-slot allowance would cover all Proposer-role traffic; a vector for "non-randao Proposer-role message 2 slots early → IGNORE" catches that.

sips/early_randao_preconsensus.md
iurii-ssv: This bullet sets a floor and nothing else, so the SIP's only concrete steer on emission timing points at the earliest legal moment — and line 20's retry rationale (attempt at `S - 2`, retry at `S - 1`) pulls defaults the same way. Line 140 prices early reveal knowability at up to 25 s against 2-4 s today and answers itself with "narrowable operationally via later emission".

Line 20 covers the reconstruction side of quorum (the latest share needed determines reconstruction time) but not its security dual: knowability also needs `2f+1`, so the exposure window tracks whatever lead a cluster's clients ship in common — the shipped default decides it, not any single operator.

Suggestion: add a RECOMMENDED lead relative to `slot_start(S)` rather than to `slot_start(S - EARLY_RANDAO_LEAD)`, keeping the 2-slot window as retry headroom. That gives "later schedules reduce the time saved" a concrete anchor and names the default that sets the network's actual exposure.

sips/early_randao_preconsensus.md
```diff
@@ -0,0 +1,137 @@
...
+- Unknown (otherwise): IGNORE; retained only if admitted under the rules below.
+
+A cached view is authoritative for a candidate iff a successfully completed proposer-duty fetch for `epoch(msg.slot)` is cached whose requested coverage includes the message's validator, evaluated against the receiver's current registry state: a fetch that predates the receiver's most recent validator-set change is not authoritative for validators affected by that change. A reorg does not revoke authority: a view made stale by a reorg remains Known, with the loss class described below. A complete-schedule fetch covers every validator, but a response without exactly `SLOTS_PER_EPOCH` distinct in-epoch slots is a fetch failure. A provisional response (one flagged execution-optimistic) installs no view. A registry-filtered fetch covers exactly the validators it requested, even when the filtered result is empty; in particular, a view covering only the operator's own validators MUST NOT yield a not-assigned verdict for claimants outside it.
```
iurii-ssv: `GET /eth/v1/validator/duties/proposer/{epoch}` takes **only** the epoch — no validator filter, in [v1](https://github.com/ethereum/beacon-APIs/blob/master/apis/validator/duties/proposer.yaml) or [v2](https://github.com/ethereum/beacon-APIs/blob/master/apis/validator/duties/proposer.v2.yaml). Every fetch returns the whole schedule; both clients narrow it afterwards — go-ssv passes `Indices` into go-eth2-client's [`ProposerDutiesOpts`](https://github.com/ssvlabs/ssv/blob/a1bcd468ff667d88c7affa58dedb3345e3467f96/beacon/goclient/proposer.go#L34), which filters client-side, and [Anchor](https://github.com/sigp/anchor/blob/unstable/anchor/duties_tracker/src/duties_tracker.rs) filters `response.data` against its `validator_indices()`.

So "requested coverage" doesn't exist at the wire: there's no registry-filtered *fetch*, only a client-side-filtered *cache*. Read literally, every receiver requested everything, its view is authoritative for every validator, and the clause's own protective last sentence never fires — a claimant its filter dropped hits line 82 (Known-and-not-assigned) → IGNORE → no retention after `d326a08` → permanent loss, the failure this clause exists to prevent. Which claimants those are depends on the filter: go-ssv's and Anchor's span the whole registry ([`Validators()`](https://github.com/ssvlabs/ssv/blob/a1bcd468ff667d88c7affa58dedb3345e3467f96/operator/duties/proposer.go#L353), [`validator_indices()`](https://github.com/sigp/anchor/blob/unstable/anchor/database/src/state.rs)), exposing only validators registered after the fetch; the own-validators view the clause itself contemplates would drop every other cluster's.

Fix, as a simplification: define authority over the **retained** schedule, and note that since every response carries all `SLOTS_PER_EPOCH` entries, retaining the slot → validator_index map (32 entries per epoch, independent of node size) makes a receiver complete-coverage by construction. Unknown then collapses to "the beacon node isn't answering".

That also drops the registry-staleness sentence — beacon assignment is registry-independent, so a full schedule already contains a validator registered after the fetch. Worth deleting rather than keeping as a stricter-than-needed option, since `d326a08` made over-approximating Unknown *unsafe*: the ePBS branch's epoch-granular [`MarkEpochsStale`](https://github.com/ssvlabs/ssv/blob/feeccd8f323486a67b6f4cf367b9c7549c10c048/operator/duties/dutystore/duties.go#L170) would leave any share arriving between an indices change and the refetch unretained, epoch-wide, not just for affected validators.

sips/early_randao_preconsensus.md
```diff
@@ -0,0 +1,137 @@
...
+
+Receivers SHOULD hold next-epoch proposer duties by the tail of each epoch.
+
+A Known view made stale by a boundary-adjacent reorg (including a provisional next-epoch view) IGNOREs an honest early share without retention. This loss class is intentional: the share is not recoverable (the in-slot re-emission is byte-identical, absorbed by the origin's publish-side deduplication and, from a restarted origin, by the receiver's seen-cache), and the cost is bounded by the round-change path, the same residual class as the round-change case in Security Considerations. Retention deliberately excludes Known-and-not-assigned; retaining it would remove the duty-fetch-failure precondition from the quarantine-exhaustion attack and would require re-evaluating the buffer on every duty-view change. Exposure is transient: receiver views converge within a slot or two of a reorg under normal operation (per-slot re-polling or dependent-root-triggered refetch).
+
+**Unknown-duty retention (IGNORE-AND-RETAIN)**
```
iurii-ssv: **This section is protocol-invisible** by its own construction: never forwarded, no penalty, no ordinary validation state mutated, "only ever consumed locally" (line 96), and promotion "cannot retroactively forward". Two clients implementing it differently — or one skipping it — are indistinguishable on the wire. Per line 85, the state it serves is a beacon-node outage, and the buffer pays off only in the sliver where the node recovers before `S`. And the DoS surface is largely self-inflicted: the ~600-validator attack at line 141 exists *because* one shared bounded buffer is mandated.

Suggest collapsing to: Unknown → IGNORE; a receiver MAY buffer such candidates locally for later promotion provided it never forwards, never penalizes, and mutates no validation state; a promoted share is revalidated as if newly received; line 106 kept verbatim. That deletes the capacity constant, the eviction order, the provisional rule, the security bullet and three vectors.

If it stays normative (shared vectors do let clients self-check, and make the line-134 measurement comparable), three spots need fixing: admission's "a refresh is already due" (95) doesn't hold for the coverage sub-case, since a cache that structurally excludes foreign validators never gets a covering refresh; the provisional rule (85 + 95) makes retention eligibility depend on unpinned upstream behavior, as go-ssv fails an optimistic node out at its health check ([`goclient.go#L518`](https://github.com/ssvlabs/ssv/blob/a1bcd468ff667d88c7affa58dedb3345e3467f96/beacon/goclient/goclient.go#L518)) and so stays eligible where a client consuming the response wouldn't; and eviction (99) breaks distance ties toward evicting the future-stamped entry, which is the one worth keeping.

sips/early_randao_preconsensus.md
iurii-ssv: Confirmed — they do expire, but with a wide margin: go-ssv's [`msgIDCacheTTL`](https://github.com/ssvlabs/ssv/blob/a1bcd468ff667d88c7affa58dedb3345e3467f96/network/topics/pubsub.go#L44) is 6.4 min (~32 slots) against this SIP's 24 s window. So "never unmark" is stronger than what implementations do; the conclusion holds, and stating the margin would be more convincing than the absolute.

sips/early_randao_preconsensus.md
shane-moore: Good catch. The previous wording compressed three separate points.

`EARLY_RANDAO_LEAD = 2` is the maximum earliness receivers must support, not an assertion that gossip needs two slots. The window gives a producer an opportunity during `S - 2` and another during `S - 1` if the earlier local publication attempt fails. Conforming clients may emit at different points in that window. Receivers accumulate shares asynchronously and reconstruct at `2f+1`, so different schedules remain compatible, but the latest share needed for quorum determines when the reveal becomes available. We did not select a longer receiver window because it would increase reveal and stale-duty exposure without an identified need.

`S >= EARLY_RANDAO_LEAD` is only a genesis underflow guard for calculating `S - 2`. It excludes absolute slots 0 and 1, not the first two slots of every epoch or the Gloas activation epoch.

Let `F` be the first slot of a non-genesis `GLOAS_FORK_EPOCH`. There is intentionally no activation warm-up:

- `S = F` is eligible, and its producer window opens at `slot_start(F - 2)`.
- `S = F + 1` is eligible, and its window opens at `slot_start(F - 1)`.
- Target slots before `F` remain ineligible.

The separate 500 ms producer delay remains guidance inside those windows. Because receiver timing also includes the clock tolerance, receiver validation and quarantine must be active no later than receiver-local `slot_start(F - 2) - EARLY_RANDAO_CLOCK_TOLERANCE`.

[Commit c859af6](https://github.com/ssvlabs/SIPs/commit/c859af659f6f3b0d9aaf386279a5f2e2c606e58e) now makes this target-slot activation model explicit in the summary, rationale, qualifying-message rules, producer behavior, and cross-client test expectations. We also aligned Anchor's [receiver](https://github.com/sigp/anchor/issues/1177), [quarantine](https://github.com/sigp/anchor/issues/1178), and [producer](https://github.com/sigp/anchor/issues/1179) issues so implementations do not accidentally introduce a current-epoch gate.

sips/early_randao_preconsensus.md
shane-moore: Answered this together with the two-slot and fork-boundary questions [here](https://github.com/ssvlabs/SIPs/pull/101#discussion_r3684450866). In short, `S >= EARLY_RANDAO_LEAD` is only the absolute-slot genesis guard; the allowed distance from the current time is enforced separately by the producer and receiver timing rules.

sips/early_randao_preconsensus.md
shane-moore: Covered this in the consolidated response [here](https://github.com/ssvlabs/SIPs/pull/101#discussion_r3684450866). We are keeping the genesis guard and not excluding the first two activation-epoch slots: `F` and `F + 1` are eligible target slots, including their pre-fork emission windows.

sips/early_randao_preconsensus.md
shane-moore: Covered this in the consolidated response [here](https://github.com/ssvlabs/SIPs/pull/101#discussion_r3684450866). There is no warm-up: target slot `F` may emit from `F - 2`, so receiver support must be active before wall-clock Gloas.

iurii-ssv (commented):
Added some potential clarifications to consider

sips/early_randao_preconsensus.md
shane-moore: Agreed. Authority now follows the retained schedule, not request coverage or the local SSV registry. A view is authoritative only when it contains exactly one duty for every slot in the epoch, and is then authoritative for every validator. This removes the requested-coverage, registry-filtering, and registry-staleness rules.

I kept `Unknown` slightly broader than "the beacon node is not answering": it means there is no complete retained schedule, which also covers startup and failed or malformed fetches.

Addressed in [a939402](https://github.com/ssvlabs/SIPs/pull/101/commits/a9394025776c09417a00003a03247818d3c82464).

sips/early_randao_preconsensus.md
shane-moore: Agreed on making retention a local `MAY`. `Unknown` now always returns `IGNORE`, and a receiver may retain the candidate or skip retention while remaining conformant. A retained candidate must first pass operator authentication and every other applicable validation rule. Initial retention cannot forward, penalize, feed signature collection or reconstruction, or mutate ordinary validation state. Promotion requires full revalidation as if newly received.

The shared capacity, deterministic eviction, provisional admission rule, attack-threshold bullet, and associated conformance vectors are removed. Capacity, eviction, expiry, and persistence are now implementation policy. The sentence about shares that are neither promoted nor normally accepted is kept verbatim.

One clarification: promotion may occur after `S` if the ordinary lateness rule still permits acceptance, since the share may still help a later round. This does not change the original `IGNORE` verdict or forwarding behavior.

Addressed in [a939402](https://github.com/ssvlabs/SIPs/pull/101/commits/a9394025776c09417a00003a03247818d3c82464).

sips/early_randao_preconsensus.md
shane-moore: Confirmed, thanks. I replaced both absolute claims, "never unmark" and "a few seconds," with the narrower statement that gossip duplicate-cache lifetimes can exceed the entire useful Early RANDAO window. The recovery wording now says the in-slot copy may be suppressed, since restart and local cache state matter.

Addressed in [a939402](https://github.com/ssvlabs/SIPs/pull/101/commits/a9394025776c09417a00003a03247818d3c82464).

sips/early_randao_preconsensus.md
shane-moore: Good catch. Addressed in [cce6fbe](https://github.com/ssvlabs/SIPs/commit/cce6fbebd5c24445470abd094232921a0266a96c): candidate classification now explicitly depends on message structure and the stamped target slot, so the duty tri-state applies to eligible in-slot and later RANDAO copies too. I also added both requested vectors: in-slot RANDAO under Unknown gets the same initial IGNORE, while non-RANDAO Proposer traffic two slots early keeps the existing timing IGNORE and does not receive the ordering exemption.

sips/early_randao_preconsensus.md
shane-moore: Agreed. I added `EARLY_RANDAO_RECOMMENDED_LEAD = SLOT_DURATION`, so an early-emitting producer SHOULD make its first publication attempt at `slot_start(S - 1)`.

The two-slot value remains the maximum receiver window. It no longer determines normal producer timing. No additional pre-slot attempt is prescribed, and the ordinary in-slot path at `S` remains the fallback.

The rationale now states that reconstruction and third-party knowability both require `2f+1` shares, so common producer policy determines typical readiness and exposure. `S - 1` seems like a reasonable initial balance: one full slot for reconstruction without defaulting to the receiver boundary. I am open to another recommended lead if you see a better tradeoff.

Addressed in [b29db38](https://github.com/ssvlabs/SIPs/commit/b29db38a8a4b073c34c3d68b24df2f881177c992).

shane-moore: You are right. The 500 ms value was additional producer-side slack, while `EARLY_RANDAO_CLOCK_TOLERANCE = 1000 ms` already handles the assumed clock disparity at the legal receiver boundary. Once the recommended emission moved to `S - 1`, the 500 ms offset no longer served a protocol purpose.

I removed it. The producer now SHOULD make its first early attempt at `slot_start(S - 1)`, while the two-slot receiver window and 1000 ms receiver tolerance remain unchanged. I am open to another recommended lead if you think there is a better choice.

Addressed in [b29db38](https://github.com/ssvlabs/SIPs/commit/b29db38a8a4b073c34c3d68b24df2f881177c992).

sips/early_randao_preconsensus.md
shane-moore: Yes, the ordinary in-slot path remains a SHOULD regardless of the recorded early outcome.

A local `published` flag is not enough to suppress it. Local enqueue, gossip publication, and peer receipt are distinct states, and successful local publication does not prove that the intended peers received the share.

The revised text also makes clear that a completed reconstruction MAY be returned immediately to the in-slot consumer without waiting for repeated signing or publication. Implementations MAY reuse cached valid signing artifacts or schedule emission independently.

Addressed in [b29db38](https://github.com/ssvlabs/SIPs/commit/b29db38a8a4b073c34c3d68b24df2f881177c992).

sips/early_randao_preconsensus.md
shane-moore: Correction to part of my earlier reply: the activation model described there is unchanged, but the two-slot value is no longer used to prescribe an `S - 2` attempt followed by an `S - 1` retry.

It remains the maximum receiver window. The recommended producer attempt is now exactly `S - 1`; the 500 ms offset and coordinated second pre-slot attempt were removed.

Addressed in [b29db38](https://github.com/ssvlabs/SIPs/commit/b29db38a8a4b073c34c3d68b24df2f881177c992).

sips/early_randao_preconsensus.md
shane-moore: Good point. A byte-identical copy shares the original content-derived gossip ID, but changing the inner partial signature and re-signing the envelope normally produces a different gossip ID. That variant reaches application validation, where the existing one-message limit for the (`MessageID`, signer, slot) key and pre-consensus or post-consensus class stops it.

I narrowed the rationale to distinguish those cases. The duplicate limit itself does not need to change.

Addressed in [c13987a](https://github.com/ssvlabs/SIPs/commit/c13987a24c73fe348416468e79a29adbb41651f8).

sips/early_randao_preconsensus.md
shane-moore: Agreed, addressed in [`8a098c9`](https://github.com/ssvlabs/SIPs/commit/8a098c9ea7c21a738d8043d7695642ae0cdd49fd).

sips/early_randao_preconsensus.md
```diff
@@ -0,0 +1,145 @@
...
+- Because gossip duplicate-cache lifetimes can exceed the entire useful Early RANDAO window, a receiver that IGNOREs an early partial cannot rely on a byte-identical in-slot re-emission to recover that share. The dedicated clock tolerance narrows timing-induced loss, and optional Unknown-duty retention can recover some missing-view cases. Remaining losses are bounded by the existing round-change path.
+
+**Specification**
+
+Notation: `S` is a proposal slot; `slot_start(s)` and `epoch(s)` as usual; for a non-genesis Gloas activation, `F` is the first slot of `GLOAS_FORK_EPOCH`; `SLOT_DURATION` is the network's beacon-chain seconds-per-slot (unchanged at Gloas, which alters only intra-slot timing). REJECT penalizes the delivering peer; IGNORE drops without forwarding or penalty.
+
+| Constant | Value | Description |
+| -------- | ----- | ----------- |
+| `EARLY_RANDAO_LEAD` | 2 slots | Maximum permitted emission lead and base receiver earliness allowance |
+| `EARLY_RANDAO_RECOMMENDED_LEAD` | `SLOT_DURATION` (12 s for a 12 s slot) | Recommended lead for an early producer's first publication attempt |
```
iurii-ssv: Unit nit: `EARLY_RANDAO_LEAD` is in slots, this one in seconds — and "12 s for a 12 s slot" reads circular. `1 slot` is dimensionally consistent; line 57 already gives the `slot_start(S - 1)` equivalence.

sips/early_randao_preconsensus.md
```diff
@@ -0,0 +1,145 @@
...
+
+Receivers SHOULD hold next-epoch proposer duties by the tail of each epoch.
+
+A retained complete schedule can become stale after a boundary-adjacent reorg while remaining Known. It can therefore yield Known-and-not-assigned and IGNORE an honest early share. This is an intentional residual: the original gossip verdict cannot be retroactively changed, a byte-identical in-slot re-emission may be duplicate-suppressed, and the cost is bounded by the existing round-change path. Receivers SHOULD refresh proposer duties promptly after detecting a change, but this SIP does not require recovery from a stale Known view.
+
+**Optional Unknown-duty retention**
+
+A receiver MAY retain an Early RANDAO candidate whose proposer-duty view is Unknown for possible later promotion. Retention is local implementation policy and does not change the gossip verdict, which remains IGNORE. A receiver that does not retain the candidate remains conformant.
```
iurii-ssv: Both round-3 reasons for excluding Known-and-not-assigned — the quarantine-exhaustion precondition and buffer re-evaluation on every view change — were properties of the mandatory shared buffer, which is gone.

Extending the MAY to any duty-assignment IGNORE would let a retaining receiver recover the stale-Known residual (96, 122, 140) through the existing promotion path, wire-invisibly: 102 already reads "other than duty assignment", and the 122 vector holds (the gossip verdict never changes). Cost: not-assigned candidates need no missing view, so supply grows — but bounding is already the implementer's job per 143.

If Unknown-only is deliberate, worth a sentence saying why.

sips/early_randao_preconsensus.md
```diff
@@ -0,0 +1,145 @@
...
+
+Before retention, the candidate MUST pass operator-signature verification and every applicable message-validation rule other than duty assignment. Initial retention MUST NOT forward the original message, penalize the delivering peer, feed the candidate to signature collection or any downstream reconstruction cache, or mutate ordinary validation state.
+
+If a retained candidate is later considered for promotion, it MUST be reprocessed as if newly received under the receiver's then-current duty view, with every stateful rule applied at that time. It may be accepted and consumed only if full revalidation succeeds; otherwise it MUST be discarded. Promotion MUST NOT retroactively forward the original message or change its original gossip verdict. An implementation may attempt promotion before or after `S`; the ordinary lateness rule determines whether the candidate can still be accepted. Capacity, keying, duplicate handling within the retention store, eviction, expiry, persistence, and earlier discard are implementation-specific.
```
iurii-ssv: The trigger is now implementation-timed ("before or after `S`"), but "otherwise it MUST be discarded" still assumes the previous install-triggered model: reprocessing under a still-Unknown view reproduces IGNORE, so a timer-driven attempt is forced to destroy the entry before any view arrives — the loss retention exists to prevent. Strictly read it's also self-contradictory, since "as if newly received" includes the Unknown path's own retention MAY (90).

Suggestion: mandatory discard only on a decided verdict — Known-and-not-assigned, or a hard failure (lateness, duplicate limit); a still-Unknown attempt leaves the entry retained. Same fix for 132's "candidates that fail revalidation are discarded".

sips/early_randao_preconsensus.md
```diff
@@ -0,0 +1,145 @@
...
+- Gossip duplicate-cache lifetimes can exceed the entire useful Early RANDAO window, so a receiver that has already seen and IGNOREd an early share cannot rely on byte-identical re-emission for recovery. A receiver restart can both clear its seen state and drop any in-memory retained candidates; restarts and partitions are correlated events. Residual: a live round-1 leader without a reconstruction forces a round change. Mitigations: later emission inside the window, unconditional in-slot emission where duplicate state does not suppress it, optional retention, and the mandatory testnet measurement above.
+- Under `EARLY_RANDAO_RECOMMENDED_LEAD`, a homogeneous cluster may make the reveal knowable close to one slot before `S`, compared with roughly 2-4 s today. The wider reaction window enables adaptive bribery, censorship, or targeted DoS when withholding the proposal favors an adversary (one randao bit per affected slot, compounding across slots). It is bounded by the existing single-proposal grinding bound and can be adjusted via producer policy without changing the receiver window.
```
iurii-ssv: This prices only the recommended default now, while boundary emission stays conformant (55) — so the previous ceiling, `EARLY_RANDAO_LEAD * SLOT_DURATION + EARLY_RANDAO_CLOCK_TOLERANCE` ≈ 25 s mainnet, remains reachable by conformant producer policy alone. Keep both numbers: ~1 slot under the default, ~2 slots as the fork-pinned ceiling.

iurii-ssv (dismissed):
Just a minor followup.

sips/early_randao_preconsensus.md
shane-moore: Agreed. I changed the constant to `1 slot` and made the slot arithmetic explicit as `slot_start(S - EARLY_RANDAO_RECOMMENDED_LEAD)`, while retaining the `slot_start(S - 1)` equivalence.

Addressed in [5cdac50](https://github.com/ssvlabs/SIPs/pull/101/commits/5cdac505c55e951315fec80a90d830711ff45e30).

sips/early_randao_preconsensus.md
shane-moore: Agreed. The Security Considerations now retain both exposure bounds: close to one slot under the recommended policy, and close to two slots if at least `2f+1` operators emit at the earliest legal boundary.

I also clarified that `EARLY_RANDAO_CLOCK_TOLERANCE` is receiver-local acceptance headroom for honest clock disparity, not additional configured producer lead.

Addressed in [5cdac50](https://github.com/ssvlabs/SIPs/pull/101/commits/5cdac505c55e951315fec80a90d830711ff45e30).

sips/early_randao_preconsensus.md
shane-moore: Agreed. I broadened the protocol-level retention `MAY` to any candidate whose only non-passing check is duty assignment, covering both Unknown and Known-and-not-assigned. This lets a receiver recover from a stale complete schedule after a replacement view arrives without changing the original `IGNORE` gossip verdict.

The larger authenticated-candidate supply is now called out in Security Considerations, while capacity, eviction, persistence, and committee scoping remain local policy. An implementation may still choose an Unknown-only subset and remain conformant.

Addressed in [8279853](https://github.com/ssvlabs/SIPs/pull/101/commits/82798537fea271d1a31d02fdc000d87b37fa3d98).

sips/early_randao_preconsensus.md
shane-moore: Agreed. I changed the revalidation rule so a retained candidate is not mandatorily discarded merely because duty assignment is still its only non-passing check. It may remain retained under local policy; full success permits promotion, while any other applicable validation failure requires discard. Expiry, eviction, and earlier local discard remain implementation-specific.

The optional-retention tests now distinguish a repeated duty-assignment-only `IGNORE` from a hard revalidation failure. Because the related change also made Known-and-not-assigned retainable, this rule applies consistently to either duty-assignment `IGNORE`.

Addressed in [8279853](https://github.com/ssvlabs/SIPs/pull/101/commits/82798537fea271d1a31d02fdc000d87b37fa3d98).

GalRogozinski (commented):
Looks good! Just a clarity request

sips/early_randao_preconsensus.md
```diff
@@ -0,0 +1,147 @@
...
...Early RANDAO message begins with the structural and canonical-form checks, which MUST run first; a structurally and canonically valid randao container satisfying the eligibility predicate is an Early RANDAO candidate, and the rules below apply to candidates only (all other messages keep today's validation unchanged). For a candidate, implementations MAY order operator-signature verification and the non-mutating contextual checks below (timing, slot ordering, duty assignment, duplicate limits) according to local denial-of-service policy, and MAY short-circuit on a contextual verdict that neither retains nor accepts the message. Any outcome that retains, accepts, forwards, feeds signature collection, or mutates ordinary validation state MUST first pass operator-signature verification. If a candidate fails both operator-signature verification and a non-retaining contextual check, either the contextual verdict or the signature-verification REJECT is conformant, but the candidate is never retained. Ordinary validation state (duplicate counts, signer state, slot high-water marks, and epoch counters) mutates only on acceptance or successful promotion, never on IGNORE or initial retention.
```
momosh-ssv: Could this lead to unbounded, unpenalized distinct-byte injection? Duplicate counts mutate only on acceptance, and under an Unknown/not-assigned view every copy is IGNOREd — so a committee operator can re-sign the envelope over arbitrarily many distinct BLS fields (unchecked at validation) for one (signer, slot). Each copy costs an RSA verification, gives no scoring feedback, and is retention-eligible with store dedup left to implementations — contradicting line 23's "bounded by the existing one-message limit". Fix: count authenticated distinct containers *seen* (REJECT the second), cap retention at one entry per key, and add a candidate-path duplicate vector (none exists).

sips/early_randao_preconsensus.md
```diff
@@ -0,0 +1,147 @@
...
+
+Candidate classification depends on message structure and the stamped target slot, not on whether a copy arrived before `slot_start(S)`. Consequently, the duty tri-state replaces existing RANDAO duty handling for every eligible target slot, including ordinary in-slot copies and later copies still within the existing Proposer lateness window.
```
momosh-ssv: What happens to relay when the duty view is Unknown? In-slot copies are now candidates, so an Unknown-view receiver IGNOREs — and stops forwarding — in-slot randao it forwards today: one node's BN gap becomes a mesh propagation hole. That makes "falls back to today's in-slot exchange" circular (the fallback hits the same tri-state that failed the early copy) and regresses never-early operators (line 64), whose shares Unknown-view peers no longer relay. Either make Unknown forward-but-don't-consume (unlike Known-not-assigned, where IGNORE is justified), or state the relay consequence, stratify the testnet measurement by duty-view state, and pin it with a vector.

sips/early_randao_preconsensus.md
```diff
@@ -0,0 +1,147 @@
...
+
+"Promoted or normally accepted" is evaluated at receipt or promotion time against the receiver's then-current duty view and is never re-evaluated at consumption. The cross-slot case is reachable two ways: a validator with two proposals in the same epoch, and a reorg that moves a proposal from `X` to `Y` after `X`-stamped shares were accepted under the pre-reorg view; the allowance exists so a completed early collection survives the shift. The signed object is identical for `X` and `Y` (`SSZUint64(epoch)`), so reuse has no cryptographic effect; implementations that key collection by signing root exercise it naturally, and implementations that require exact-slot consumption remain conformant.
```
momosh-ssv: Do we need a per-signer dedup rule? In this duty-move case receivers hold two accepted shares from one signer for the epoch (X- and Y-stamped, identical inner signature and root). A root-keyed collector counting entries toward 2f+1 hits false quorum — 4 ops, f=1: 2 signers × 2 stamps = 4 entries vs threshold 3 — and reconstructs an invalid reveal that fails only at the BN. One sentence fixes it: a reconstruction counts at most one share per signer per signing root (a second replaces, never accumulates); the duty-move vector can then assert 2f+1 distinct signers over a mixed X/Y set.

sips/early_randao_preconsensus.md
```diff
@@ -0,0 +1,147 @@
...
+
+Candidate classification depends on message structure and the stamped target slot, not on whether a copy arrived before `slot_start(S)`. Consequently, the duty tri-state replaces existing RANDAO duty handling for every eligible target slot, including ordinary in-slot copies and later copies still within the existing Proposer lateness window.
+
+Checks are staged: operator-signature verification gates retention, acceptance, forwarding, and state mutation; BLS-share validity gates consumption. A candidate is fully qualifying only once every applicable check has passed; honest producers emit qualifying messages by construction.
+
+*Earliness.* A receiver MUST accept a candidate's timing iff
```
momosh-ssv: Should we consider bounding only the early side? "MUST accept a candidate's timing iff" holds for any already-started slot — even ten epochs late — so one conformant reading has the formula override the lateness TTL and accept/forward arbitrarily late randao, leaving the past-slot half of the ordering exemption unbounded. "MUST NOT reject as too early iff …", with lateness applying independently, removes the ambiguity; the vectors cover only the early boundary, so add a late one.

sips/early_randao_preconsensus.md
```diff
@@ -0,0 +1,147 @@
...
+
+```text
+slot_start(msg.slot) - local_now <= EARLY_RANDAO_LEAD * SLOT_DURATION + EARLY_RANDAO_CLOCK_TOLERANCE
+```
+
+inclusive on the accept side; strictly greater is IGNORE. This replaces the generic clock tolerance for this rule only. Lateness is unchanged (existing Proposer-role TTL against `msg.slot`). All other message classes keep their existing allowances.
+
+`EARLY_RANDAO_CLOCK_TOLERANCE` is receiver-side headroom covering the assumed maximum 1 s pairwise honest clock disparity at the producer's legal boundary. It does not shift the recommended producer time.
+
+*Slot ordering.* Candidate randao partials are exempt from the per-(operator, `MessageID`) highest-seen-slot rule in both directions: they MUST NOT be rejected for a slot below the high-water mark, and they MUST NOT advance the high-water mark applied to other message classes on the same `MessageID` (randao shares its `MessageID` with the proposal's consensus and post-consensus traffic). Final acceptance still requires the remaining qualifying-message checks; earliness, lateness, the duplicate limit of 1, and canonical form bound these messages instead.
```
momosh-ssv: Any reason "slot high-water marks" is still in line 68's mutate-on-acceptance list? It contradicts the MUST NOT here: an early candidate for slot 11 accepted during slot 9 either advances the mark — rejecting still-live slot-10 consensus/post-consensus traffic on the same MessageID under the unchanged rule — or it doesn't, and implementations diverge. State that candidates neither read nor write any high-water mark (scope line 68's list to non-candidates). The consecutive-slots vector is randao-vs-randao only; this cross-class case is what the MUST NOT actually regulates.

sips/early_randao_preconsensus.md
```diff
@@ -0,0 +1,147 @@
...
+- Known and not assigned: the duty-assignment verdict is IGNORE, never REJECT (receiver views can be stale or re-orged); an independently failing rule may still determine the final verdict under the ordering allowance above.
+- Unknown (otherwise): IGNORE.
+
+Either duty-assignment IGNORE is eligible for the optional local retention below.
+
+A proposer-duty view is authoritative for an epoch iff the receiver retains a complete schedule containing exactly `SLOTS_PER_EPOCH` duties, one for every distinct slot in that epoch. Such a view is authoritative for every validator in the epoch, independent of the receiver's local SSV registry. A schedule that fails this shape establishes no new view. A newly retained complete schedule replaces the prior view; a failed or malformed refresh and a local validator-set change do not invalidate a previously retained authoritative view. A reorg likewise does not revoke authority until a replacement complete schedule is retained, so a stale view remains Known with the loss class described below.
+
+Receivers SHOULD hold next-epoch proposer duties by the tail of each epoch.
```
momosh-ssv: Might be worth tightening this SHOULD? For S in the first EARLY_RANDAO_LEAD slots of an epoch, the whole emission window is in the prior epoch, so every relay hop needs E+1's complete schedule — and "tail of each epoch" is undefined. With IGNORE-on-Unknown gating relay, the early path degrades network-wide for 2 of every 32 slots, exactly where duplicate suppression then blocks the in-slot copy. Suggest a MUST with a deadline (e.g. hold E+1 by slot_start(last_slot(E) - EARLY_RANDAO_LEAD)) plus a first-two-slots vector.

sips/early_randao_preconsensus.md
```diff
@@ -0,0 +1,147 @@
...
+
+- Known (a retained view is authoritative for the candidate; see below) and assigned at exactly `msg.slot`: the duty-assignment check passes.
+- Known and not assigned: the duty-assignment verdict is IGNORE, never REJECT (receiver views can be stale or re-orged); an independently failing rule may still determine the final verdict under the ordering allowance above.
+- Unknown (otherwise): IGNORE.
+
+Either duty-assignment IGNORE is eligible for the optional local retention below.
+
+A proposer-duty view is authoritative for an epoch iff the receiver retains a complete schedule containing exactly `SLOTS_PER_EPOCH` duties, one for every distinct slot in that epoch. Such a view is authoritative for every validator in the epoch, independent of the receiver's local SSV registry. A schedule that fails this shape establishes no new view. A newly retained complete schedule replaces the prior view; a failed or malformed refresh and a local validator-set change do not invalidate a previously retained authoritative view. A reorg likewise does not revoke authority until a replacement complete schedule is retained, so a stale view remains Known with the loss class described below.
```
momosh-ssv: What happens if the epoch is 0? duties/proposer/0 returns 31 entries (the genesis slot has no proposer), so "exactly SLOTS_PER_EPOCH duties" is unsatisfiable: every epoch-0 schedule establishes no view and all epoch-0 candidates stay Unknown on the GLOAS_FORK_EPOCH = 0 network the vectors contemplate. Define completeness as one duty per slot that has a proposer assignment, or exempt slot 0 of epoch 0 — worth confirming client behavior first.

sips/early_randao_preconsensus.md
```diff
@@ -0,0 +1,147 @@
...
+```
+
+inclusive on the accept side; strictly greater is IGNORE. This replaces the generic clock tolerance for this rule only. Lateness is unchanged (existing Proposer-role TTL against `msg.slot`). All other message classes keep their existing allowances.
+
+`EARLY_RANDAO_CLOCK_TOLERANCE` is receiver-side headroom covering the assumed maximum 1 s pairwise honest clock disparity at the producer's legal boundary. It does not shift the recommended producer time.
+
+*Slot ordering.* Candidate randao partials are exempt from the per-(operator, `MessageID`) highest-seen-slot rule in both directions: they MUST NOT be rejected for a slot below the high-water mark, and they MUST NOT advance the high-water mark applied to other message classes on the same `MessageID` (randao shares its `MessageID` with the proposal's consensus and post-consensus traffic). Final acceptance still requires the remaining qualifying-message checks; earliness, lateness, the duplicate limit of 1, and canonical form bound these messages instead.
+
+*Duty assignment.* Tri-state, evaluated per candidate against the receiver's proposer-duty view for `epoch(msg.slot)` and the message's validator:
```
momosh-ssv: Do we need to handle a validator that can't be resolved to an index? Schedules key by validator_index, the MessageID by pubkey — so even with an authoritative view a just-registered validator is neither assigned nor provably not-assigned, and no branch is specified. It matters: pre-existing unknown-validator checks may REJECT (penalizing an honest share) where the tri-state says IGNORE, and retention eligibility differs (duty assignment must be the only failing check). State precedence — existing rules resolve first with IGNORE not REJECT for candidates, or classify as Unknown — and narrow line 94's registry-independence claim to schedule authority.

sips/early_randao_preconsensus.md
```diff
@@ -0,0 +1,147 @@
...
+
+**Producer behavior**
+
+For a locally known proposer duty at eligible slot `S`, an operator:
+
+- MAY broadcast its qualifying randao partial at any wall-clock time in `[slot_start(S - EARLY_RANDAO_LEAD), slot_start(S))`;
+- MUST NOT broadcast it before `slot_start(S - EARLY_RANDAO_LEAD)` by its own clock;
+- if it elects to emit early and knows the duty by the recommended time, SHOULD make its first local publication attempt at `slot_start(S - EARLY_RANDAO_RECOMMENDED_LEAD)`, equivalent to `slot_start(S - 1)`;
```
momosh-ssv: Worth considering an intra-slot offset? slot_start(S - 1) makes every early emitter fire at a slot boundary, competing with S-1 block propagation and the compressed post-Gloas attestation deadline. slot_start(S - 1) + SLOT_DURATION/2 still leaves ~6 s for gossip and reconstruction, halves the reveal-exposure window, and avoids the burst; at minimum add intra-slot offset to the mandatory lead comparison.

sips/early_randao_preconsensus.md
```diff
@@ -0,0 +1,147 @@
...
+
+*Earliness.* A receiver MUST accept a candidate's timing iff
+
+```text
+slot_start(msg.slot) - local_now <= EARLY_RANDAO_LEAD * SLOT_DURATION + EARLY_RANDAO_CLOCK_TOLERANCE
+```
+
+inclusive on the accept side; strictly greater is IGNORE. This replaces the generic clock tolerance for this rule only. Lateness is unchanged (existing Proposer-role TTL against `msg.slot`). All other message classes keep their existing allowances.
+
+`EARLY_RANDAO_CLOCK_TOLERANCE` is receiver-side headroom covering the assumed maximum 1 s pairwise honest clock disparity at the producer's legal boundary. It does not shift the recommended producer time.
```
momosh-ssv: Should the tolerance match how the bound is stated? "1 s pairwise" is exactly covered by 1000 ms — but clock bounds are usually per node vs true time, and ±1 s per node gives up to 2 s producer-receiver skew, IGNOREing honest boundary emissions. Pin one: ±500 ms per node, or tolerance = twice the per-node bound. Only boundary emitters are exposed (recommended-lead producers have a full slot of slack) — worth saying.

sips/early_randao_preconsensus.md
```diff
@@ -0,0 +1,147 @@
...
+
+Producer implementations that emit early SHOULD test the recommended `S - 1` attempt, an injected local publication failure followed by the in-slot path, duty discovery after the recommended time, unconditional invocation of the in-slot path with and without a completed reconstruction, and completion of a cached in-slot consumer without waiting for repeated signing or publication.
+
+Before mainnet activation guidance, testnet measurement MUST quantify early-share miss incidence at duty start (stratified by emission lead, receiver restarts, partition duration, leader round, and whether receiver retention was enabled), record promotion timing and outcome where retention is enabled, and MUST demonstrate round-2 viability under Gloas timing, meaning block publication before the post-Gloas useful-block deadline. The emission-lead comparison MUST include `EARLY_RANDAO_RECOMMENDED_LEAD`, at least one earlier lead, and at least one later lead, and MUST record threshold-ready time relative to `slot_start(S)` so the resulting reveal-exposure window is explicit.
```
momosh-ssv: Any reason "the post-Gloas useful-block deadline" isn't pinned to a definition? It sits inside a MUST that gates mainnet activation guidance, so it needs a citable value or a cross-reference into the ePBS SIP (#94) — the same treatment GLOAS_FORK_EPOCH got.

sips/early_randao_preconsensus.md
```diff
@@ -0,0 +1,147 @@
...
+
+**Motivation**
+
+Gloas moves the attestation deadline to 1/4 slot. RANDAO pre-consensus (sign, gossip, collect 2f+1, reconstruct) is the first serial step of block production and costs a gossip round trip at slot start. The signed object, `SSZUint64(epoch)` under `DOMAIN_RANDAO`, is a pure function of the proposal slot's epoch, so it can be signed and exchanged at any time and a re-org can never invalidate it. Failure modes fall back to today's in-slot exchange, except at a receiver that already IGNOREd the early copy and did not retain it: gossip duplicate caching may suppress the byte-identical in-slot copy for the remainder of the useful duty window (see *Duty assignment* and *Optional duty-assignment retention*). Addresses [ssv-spec#373](https://github.com/ssvlabs/ssv-spec/issues/373).
```
momosh-ssv: Might be worth softening "a re-org can never invalidate it"? A reorg can move the validator's assignment out of the epoch entirely — which is exactly why the reuse allowance in Reconstruction and consumption is epoch-scoped. "A re-org within the epoch can never invalidate it" matches the Specification.

MatheusFranco99:
Following up on the message validation duty-staleness design.

I initially thought accepting a message under a missing duty view made no sense — without the duty you can't process it. But that reasoning has two issues: most nodes validating a message are relays that don't actually process it (non-committee), and on a committee node an accepted message just waits in the queue until the duty fetch lands. We already do exactly this on purpose btw in ssv node: the first-slot-of-epoch carve-out for RANDAO messages accepts while duties "are still being fetched." Meanwhile IGNORE is not as harmless as gossipsub's redundancy suggests, because the staleness causes here may be correlated (e.g. a common BN that impacts two operators).

On the other hand, accepting cannot be as simple to prevent spamming.
So my proposal for the proposer role is a tri-state:

- Known beacon state and assigned → pass.
- Known and not assigned → IGNORE (you have real evidence; receiver views can be stale, so never REJECT).
- Unknown → accept, backstopped by a new per-epoch duty cap for the proposer role (it's the one role missing from `dutyLimit` today).

Two honest caveats. First, unlike the registration/aggregator caps, a proposer cap is probabilistic rather than structural — a validator *can* draw more than 2 proposals in an epoch, it's just vanishingly rare — so the rule should state that, and the value (2, or 4 to be generous) is a judgment call (I think it's fine increasing up to the max possible, only for the proposer role ofc). Also, the cap fallback only ever binds while the view is Unknown (and soon we will have EIP-7917). Second, forwarding under Unknown is safe reputation-wise: duty-assignment failures are IGNORE-class everywhere, and every REJECT-class check (signatures, structure, counts) needs no duty view, so we never relay something a peer would penalize us for.

One consequence worth making explicit: this cap removes the stated justification for the early-RANDAO rule's fail-closed Unknown ("the assignment check is the only spam bound"). With the cap in place, the same tri-state can govern the RANDAO share and the legacy QBFT/post-consensus path — resolving the asymmetry between the two within one proposer flow, rather than documenting it as intentional. And EIP-7917's in-state proposer lookahead makes the Known side stronger over time: a complete schedule two epochs ahead, pinned by a dependent root, shrinks Unknown to deep-reorg and outage windows.

